PR #25, opened 2019-05-01
:ambulance: author of events need to have a refresh like that numbers…

### LINKED ISSUE #19: Always send refresh for author refresh in web hook
Add default :

```go
return []*HandleWebhook{{From: senderGitlabUsername}}, nil
```

to all web hook handler to refresh left summary compo for author

manland:
… in webapp are always true

Fix #19

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ The changelog can be found at https://github.com/mattermost/mattermost-plugin-gi
 
 ## 0.2.0 - unreleased
 
+- Send refresh to webapp of the author of events received by webhook [#19](https://github.com/manland/mattermost-plugin-gitlab/pull/19)
 - Add all webhook events for pipeline (run, fail, success) [#17](https://github.com/manland/mattermost-plugin-gitlab/pull/17)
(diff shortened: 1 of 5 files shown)


manland:
@hanzei thanks for motivation I learn a lot to make this test :D I think we need to add an example of package `mattermost-server/plugin/plugintest` in the sample-plugin project!

If you have tips with mocks, please tell me, its the first time I use it. In my others projects I prefer to use [IoC pattern](https://en.wikipedia.org/wiki/Inversion_of_control) like I do in `webhook` package. But every day I dev in java (kotlin) and we use lot of mocks... I need to try myself to use them in golang :D

hanzei:
Glad you learned a lot of this. BTW: The demo plugin has a lot of tests, which uses  mocks a lot. See https://github.com/mattermost/mattermost-plugin-demo/blob/master/server/activate_hooks_test.go for example. On thing I find very useful is using a  `SetupAPI func() *plugintest.API` field in table tests. (https://github.com/mattermost/mattermost-plugin-demo/blob/master/server/activate_hooks_test.go#L20)

PR #43, opened 2019-05-11
:art: rename EntrepriseBaseUrl config to GitlabURL

### LINKED ISSUE #34: Setting  `Enterprise Base URL` is not appropriate for GitLab
This property is copied from `mattermost-plugin-github` and because in github all on-premise instances are entreprise grade, the name is good. But gitlab can have on-premise without etreprise lincence.

I propose to rename it `Gitlab URL` with `https://gitlab.com` as default value, without optional.

manland:
Fix #34

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -19,7 +19,7 @@
         "settings": [
             {
                 "key": "GitlabURL",
-                "display_name": "Gitlab Base URL",
+                "display_name": "Gitlab URL",
                 "type": "text",
                 "default": "https://gitlab.com",
                 "help_text": "The base URL for using the plugin with a GitLab installation. Examples: https://gitlab.com or https://gitlab.example.com"
--- a/plugin.json
+++ b/plugin.json
@@ -18,6 +18,13 @@
         "header": "To set up the GitLab plugin, you need to register a GitLab OAuth app here https://gitlab.com/profile/applications.",
         "settings": [
             {
+                "key": "GitlabURL",
+                "display_name": "Gitlab Base URL",
+                "type": "text",
+                "default": "https://gitlab.com",
+                "help_text": "The base URL for using the plugin with a GitLab installation. Examples: https://gitlab.com or https://gitlab.example.com"
+            },
+            {
                 "key": "GitlabOAuthClientID",
                 "display_name": "GitLab OAuth Client ID",
                 "type": "text",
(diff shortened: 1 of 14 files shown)


plugin.json
```diff
@@ -16,6 +16,13 @@
     "settings_schema": {
         "header": "To set up the GitLab plugin, you need to register a GitLab OAuth app here https://gitlab.com/profile/applications.",
         "settings": [
+            {
+                "key": "GitlabURL",
+                "display_name": "Gitlab Base URL",
+                "type": "text",
+                "default": "https://gitlab.com",
```
hanzei: :+1:

plugin.json
```diff
@@ -16,6 +16,13 @@
     "settings_schema": {
         "header": "To set up the GitLab plugin, you need to register a GitLab OAuth app here https://gitlab.com/profile/applications.",
         "settings": [
+            {
+                "key": "GitlabURL",
+                "display_name": "Gitlab Base URL",
```
hanzei: Maybe `Gitlab URL` instead of `Gitlab Base URL`? 1/5

server/configuration.go
```diff
@@ -42,6 +43,9 @@ func (c *configuration) Clone() *configuration {
 
 // IsValid checks if all needed fields are set.
 func (c *configuration) IsValid() error {
+	if _, err := url.ParseRequestURI(c.GitlabURL); err != nil {
```
hanzei: Why `url.ParseRequestURI` instead of `url.Parse`?

server/configuration.go
```diff
@@ -42,6 +43,9 @@ func (c *configuration) Clone() *configuration {
...
+		return fmt.Errorf("Must have a valid GitLab URL")
```
hanzei: Nit: 
```suggestion
		return errors.New("Must have a valid GitLab URL")
```

server/configuration.go
manland: `url.Parse` can be [relative](https://godoc.org/net/url#Parse) : `/toto/tata` is valid

plugin.json
manland: :heavy_check_mark:

manland:
Again... thank you very much @hanzei to take time to review all my PR :100: :1st_place_medal:

PR #53, opened 2019-05-21
WIP: :sparkles: add webhook in gitlab when subscribe

### LINKED ISSUE #15: Add webhook without leaving mattermost
As a user (before I dev this plugin) I didn't know how to add a project to listen to in github-plugin. I didn't read the readme... meaculpa. But I think to be more easy to use we should add webhook for user. Like that user just `/gitlab subscribe owner/project` and all the magic make the job.

Obviously for group, we need to fetch all projects and add it to every repos. If user create a new project after this init step we will miss it... maybe we can cron a synchronizer... 

https://docs.gitlab.com/ce/api/projects.html#add-project-hook

### LINKED ISSUE #46: Help me with overstanding the ReadMe steps...
Unable to find this part nor any identical settings as described here:

I just realized now this is the old readme that applies to github mostly maybe thats why I do not find all menus like described ** updated **

**```
**Go to the settings page of your GitLab group and click on "Webhooks" in the sidebar
Click "Add webhook"** WHERE is this area Gitlab Group??? <=

Change content type to "application/json" <= Where is this? Maybe adding a Print screen visual or GIF or short video, very confusing this readme .

Select the events: Issues, Issue comments, Pull requests, Pull request reviews, Pull request review comments, Pushes, Branch or Tag creation and Branch or Tag deletion

image

I am curious where are your events taken from... Again confusing here while I could never match all events you described.

```**

### PR DESCRIPTION
**Pull request by manland**
TODO : 

* [ ] test
* [ ] all pages in `addWebhook` to check already exist

Fix #15 
Fix #46

### CODE REVIEW

**Comment by manland** [comment]
@hanzei @crspeller @jasonblais I need help for this... I don't know if I finish it or not!

Hi think, adding webhook automatically, is a good feature. But I don't know if this MR is not over-engineering for the purpose.

I can imagine a simple one where user add an arg to the command `/gitlab subscribe owner[/repo] [features] autowebhook=true` and it will add the webhook only for `owner/repo` not for `owner` at the user demand (not automatically)!?

In this PR I try to keep `owner` (or organization in github world) and sub project in sync (every time we create a project in an organization and plugin listen this organization we add a webhook).

And what about the unsucribing a repo? I need to delete the webhook? Or I keep it running (to have info on issues... for DM user) and just don't display it in the channel?

An other missing feature around this topic is #36, I can imagine a popup with all subscriptions and a form to add new one. And so adding a check box to add webhook on creation...

What do you think about that? Finishing it, is not time consuming but can have huge impact for user!? Maybe a more simple implementation and try to have feedback from users can help?

File: server/cron.go [hunk 1]
```diff
@@ -0,0 +1,56 @@
...
+	p *Plugin
+	c *cron.Cron
+}
+
+// NewCron return a cron
+func NewCron(p *Plugin) (*Cron, error) {
+	c := cron.New()
+	config := p.getConfiguration()
+
+	if err := c.AddFunc("@every 1h", func() {
```
manland: This is the part i'm afraid of : evry hour for all subscriptions check if a web hook is running, if not create one... I can improve it by adding info in KVStore and just ask for organization with new project. But if a user has huge subscriptions it will be time consuming and potentially generate lot of traffic...

jasonblais:
@manland To confirm - is the proposal to automate webhook creation [in step 3](https://github.com/mattermost/mattermost-plugin-gitlab#installation) based on users' subscriptions?

guillaume-perreal:
User here: if an administrator of a team could use "/gitlab subscribe group/subgroup/project" to create the webhook in said Gitlab project, given said administrator also has the required permissions, you would already have a way better workflow than the current one. This could be a feature release on its own.

Handling whole (sub)group subscriptions could come in a second time/PR.

hanzei:
This PR is not longer needed

PR #54, opened 2019-05-21
:memo: clean configuration for plugin

### LINKED ISSUE #51: Add placeholder text to Gitlab group field in System Console
On the System Console page, we should add an example to the Gitlab group field? This would make it clear if the field is expecting the URL to the group or simply the name.

manland:
Fix #51


manland:
Tanks @hanzei :+1:

PR #55, opened 2019-05-21
:bug: user not logged can see help command

### LINKED ISSUE #49: Help command should be available before the user is connected
Help for the command is provided in the form of a DM from the BOT once connected but, it would also be good to have /gitlab help available to the user before they authenticate to show connect and the other commands.

manland:
Fix #49


server/command.go
```diff
@@ -77,6 +77,9 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 
 		resp := p.getCommandResponse(model.COMMAND_RESPONSE_TYPE_EPHEMERAL, fmt.Sprintf("[Click here to link your GitLab account.](%s/plugins/%s/oauth/connect)", *config.ServiceSettings.SiteURL, manifest.Id))
 		return resp, nil
+	} else if action == "help" || action == "" {
```
hanzei: Nit: You can drop the `else`

manland: :+1:

manland:
Again and again very thank yu @hanzei !

PR #73, opened 2019-07-01
Unify build tools with mattermost-plugin-starter-template

### LINKED ISSUE #71: Caching in ci doesn't work
On issue with this setup is that the caching doesn't work. The orb expects the `go.mod` file to be at `/go.mod`. But in this project it's at `/server/go.mod`.
I see 
Screenshot from 2019-06-24 07-45-29

Possible options:
1. Ignore this problem
2. Move `go.mod` into the root directory
3. Modify the orb to make this customizable

@manland Thoughts on this?

_Originally posted by @hanzei in https://github.com/mattermost/mattermost-plugin-gitlab/pull/69#issuecomment-504924341_

hanzei:
This ended up being are more holistic change. I streamlined the entire tool chain to be in sync with mattermost-plugin-starter-template. This allows for easier updates in the future.

Fixes #71


Makefile
```diff
@@ -23,7 +31,7 @@ apply:
 
 ## Runs govet and gofmt against all packages.
 .PHONY: check-style
-check-style: webapp/node_modules gofmt govet errcheck
+check-style: webapp/.npminstall gofmt govet errcheck
```
hanzei: "Normally" `golint` would be part of this target. But given that the code isn't compliant (jet) I leaved this out.

manland: Well done the add of golint! I have missed the addition in default plugins workflow but I :+1: now!

hanzei: There are currently 57 error when running `golint`. Fixing them will take some time :wink:

PR #82, opened 2019-08-09
List commits on push event

hanzei:
I don't have a local GitLab instance to test this.


manland:
It seams cool, thank you !

What about adding a test for commit**s** to be sur all is ok?

hanzei:
Sure, done

PR #84, opened 2019-08-09
Link to root URL with button

hanzei:
There is no value in linking to the `/oauth/applications/` page, once the user is connected.


manland:
It was to mimic the github plugin which link to the application linked into the github app. But in gitlab there is no direct link so I have targeted the most precise link!

Anyway, your right it's better like that ;)

PR #106, opened 2019-09-29
decrease size of new issue messages

svelle:
The new issue notifications are way too big. IMO I'd even go with three `#`'s but two should be fine as well.


svelle:
Just a heads up I might rework this for all messages. Should I then repurpose this PR or close and open a new one?

hanzei:
Would you mind opening a new PR?

svelle:
@hanzei absolutely no problem.

PR #119, opened 2019-11-20
MM-20306 - Fixing header button spacing

asaadmahmood:
Ticket: https://mattermost.atlassian.net/browse/MM-20306

Screenshot 2019-11-20 at 7 07 06 PM


hanzei:
@asaadmahmood Would you please also fix this upstream? https://github.com/mattermost/mattermost-plugin-github/blob/master/webapp/src/components/sidebar_buttons/sidebar_buttons.jsx#L217-L226

PR #120, opened 2019-11-21
Fixed wording when unable to find project

icelander:


manland:
Thank you @icelander :rocket:

Issue #74, opened 2019-07-04
Configuration change must propagate to client side


manland:
## Reproduction

* I activate the plugin, default config point to gitlab.com
* A user load MM
* I configure it with my.gilab.com
* User connect to my.gitlab.com but top/left widget point to gitlab.com

## Fix

When OnConfigurationChange in server arrive we must prevent all user connected (web socket) to reload conf

manland:
It seams to be on each first connect.

* User is not logged
* User log
* All icons are linked to gitlab.com instead of private instance
* After f5 links are good and point to private instance

larkox:
May I take this one? First time contributing to plugins, so I may need some extra help 😊

manland:
@larkox yes of course go for it! And ask any question you have here or on https://community.mattermost.com (channel plugin-gitlab or developer-toolkit) ;)

PR #122, opened 2019-12-02
[GH-74] Configuration change must propagate to client side

larkox:
Summary:
Fix "Configuration change must propagate to client side"

Link:
Fixes https://github.com/mattermost/mattermost-plugin-gitlab/issues/74

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -206,6 +206,8 @@ func (p *Plugin) completeConnectUserToGitlab(w http.ResponseWriter, r *http.Requ
 			"connected":        true,
 			"gitlab_username":  userInfo.GitlabUsername,
 			"gitlab_client_id": config.GitlabOAuthClientID,
+			"gitlab_url":       config.GitlabURL,
+			"organization":     config.GitlabGroup,
 		},
 		&model.WebsocketBroadcast{UserId: userID},
 	)


manland:
@larkox Thanks for this contribution!

@hanzei do you know why codecov is failing? @larkox has asked in [community](https://community.mattermost.com/core/pl/3awkmnk5yty7jja8s4sdo1p61o) without answer. I think it's because coverage is low and in this case you can't do anything @larkox ;)

@hanzei do you think someone in MM can test this dev? In plugin environment?

hanzei:
Codecov ist not failing. It just complains that the newly added code is not covered. It's an optional check that you can ignore.

hanzei:
@DHaussermann Could you help with testing this?

DHaussermann:
@manland or @larkox  I'm not sure I understand the issue here. I'm not seeing a change. Maybe the way I'm testing this is not valid?

Steps:
- GitLab URL is set to `https://gitlab.com`
- In Chrome browser connect a user to GitLab
- As admin in a different session update GitLab URL to `https://gitlab.com/dhaussermann` and Save
**Expected**: Connected user is other browser session see's GitLab icon in LHS pointed to current URL 
**Observed**: Connected user is other browser session see's GitLab icon in LHS pointed to previous URL until they hit refresh in the browser.

larkox:
The original problem is that if you have a custom URL, when you connect, it always points to ```gitlab.com``` until you reload the page. With this fix, it has the correct URL when you connect.

PR #137, opened 2020-02-28
[GH-15] manage webhooks with slash commands

ericjaystevens:
#### Summary
Add features to manage projects and group scoped GitLab webhooks.

```
/gitlab webhook list group[/project]
/gitlab webhook add group[/project] [url] [triggers] [token]-- if url is omitted points to plugin URL, if triggers is blank sets all, if token is omitted uses plugin's token
```

When a user subscribes to a new project or group, it will tell them the command to run to create a webhook, if one isn't already created for the group/project.

### TODO

- [x] list project hooks
- [x] list group hooks
- [x] create project hooks
- [x] create group hooks  

#### Ticket Link

fixes https://github.com/mattermost/mattermost-plugin-gitlab/issues/15
fixes https://github.com/mattermost/mattermost-plugin-gitlab/issues/156


ericjaystevens:
I've made all the required upstream changes, and it was just released today https://github.com/xanzy/go-gitlab/releases/tag/v0.30.0

a little more work on my end and this should be ready to move forward.

ericjaystevens:
feature complete, @hanzei Work in Progress can be removed.

hanzei:
@ericjaystevens  Heads up that there is a merge conflict to resolve

server/command.go
```diff
@@ -184,8 +214,179 @@ func (p *Plugin) ExecuteCommand(_ *plugin.Context, args *model.CommandArgs) (*mo
...
+			formatedWebhooks += hook.Stringify()
+		}
+		return formatedWebhooks
+
+	case "add":
+		namespace := parameters[1]
+		fullPath := strings.Split(namespace, "/")
+
+		siteURL := *p.API.GetConfig().ServiceSettings.SiteURL
+		urlPath := fmt.Sprintf("%v/%s", siteURL, inboundWebhookURL)
```
iomodo, line 281: Shouldn't we add the port as well in case it is not `:80`?

ericjaystevens: Good Thinking, I use port 8065 in my test environment. But I have the port number included in my site URL.

Looking at the System Console under site URL I see, 

> The URL that users will use to access Mattermost. Standard ports, such as 80 and 443, can be omitted, but non-standard ports are required. For example: http://example.com:8065. This setting is required.

So I think it might be safe to assume the port will be included in the Site URL. But if that's not a safe assumption, it should be an easy check.

iomodo: You are absolutely right. One case worth checking might be `siteURL` being empty, because it's a default. I'd suggest using 

`"localhost" + *p.API.GetConfig().ServiceSettings.ListenAddress`

ericjaystevens: I think it's a good idea to check siteURL for empty, I'll add that. 

I'm pretty sure other things won't work (like subscribe and activate) if siteURL is not set. But if gets unset webhook creation would set webhooks to point to just the inboundWebhookURL, which would be unexpected.

I'm not sure computing the URL is practical or desired, they could be behind a load balancer or proxy and we wouldn't know their external IP/FQDN. 

How about I just add a check, and if it is empty I fail the webhook creation with a message like

> Unable to create webhook. The Mattermot Site URL is not set. Set it in the admin console or rerun ```/gitlab webhook add group/project URL``` including the desired URL.

server/command.go
```diff
@@ -31,9 +31,41 @@ const COMMAND_HELP = `* |/gitlab connect| - Connect your Mattermost account to y
...
+	 * Comments 
+	 * ConfidentialComments 
+	 * IssuesEvents
+	 * ConfidentialIssuesEvents 
+	 * MergeRequestsEvents 
+	 * JobEvents 
+	 * PipelineEvents 
+	 * WikiPageEvents
+	 * SSLverification
+  * |url| is the URL that will be called when triggered. Defaults to this plugins URL
```
hanzei, line 51: What is a use case for setting this not to the plugin URL?

server/command.go
```diff
@@ -31,9 +31,41 @@ const COMMAND_HELP = `* |/gitlab connect| - Connect your Mattermost account to y
...
+  * |token| Secret token. Defaults to secrete token used in plugin's settings.
```
hanzei, line 52: Similar question here

server/command.go
```diff
@@ -31,9 +31,41 @@ const COMMAND_HELP = `* |/gitlab connect| - Connect your Mattermost account to y
...
+`
 const webhookHowToURL = "https://github.com/mattermost/mattermost-plugin-gitlab#step-3-create-a-gitlab-webhook"
+const inboundWebhookURL = "plugins/com.github.manland.mattermost-plugin-gitlab/webhook"
+const unknownActionMessage = "Unknown action, please use `/gitlab help` to see all actions available."
+const newWebhookEmptySiteURLmessage = "Unable to create webhook. The Mattermot Site URL is not set. " +
+	"Set it in the Admin Console or rerun /gitlab webhook add group/project URL including the desired URL."
```
hanzei, line 58: Nit pick:
```suggestion
const (
	webhookHowToURL = "https://github.com/mattermost/mattermost-plugin-gitlab#step-3-create-a-gitlab-webhook"
	inboundWebhookURL = "plugins/com.github.manland.mattermost-plugin-gitlab/webhook"
	unknownActionMessage = "Unknown action, please use `/gitlab help` to see all actions available."
	newWebhookEmptySiteURLmessage = "Unable to create webhook. The Mattermot Site URL is not set. " +
	"Set it in the Admin Console or rerun /gitlab webhook add group/project URL including the desired URL."
)
```

server/command.go
```diff
@@ -246,27 +453,34 @@ func (p *Plugin) subscribeCommand(parameters []string, channelID string, config
...
+		return fmt.Sprintf("Successfully subscribed to %s.", fullPath)
 	}
+	//no web hook found
+	hookStatusMessage = fmt.Sprintf(
+		"\nA Webhook is needed, run ```/gitlab webhook add %s``` to create one now.",
+		fullPath,
+	)
 
-	return fmt.Sprintf("Successfully subscribed to %s. %s", fullPath, hookStatusMessage)
+	return fmt.Sprintf("Successfully subscribed to %s.%s", fullPath, hookStatusMessage)
```
hanzei, line 485: Nit pick: I would be nice if the message `Successfully subscribed to` would not have to be duplicated in the code.

server/gitlab/api.go
```diff
@@ -11,16 +11,198 @@ import (
...
+	var webhooks []*WebhookInfo
+	for _, hook := range hooks {
+		webhooks = append(webhooks, getGroupHookInfo(hook))
+	}
+
+	return webhooks, nil
+}
+
+//Stringify produces a multiline bulleted string for displaying webhook information.
+func (w *WebhookInfo) Stringify() string {
```
hanzei, line 106: `fmt` definer a `Stringer` interface with one only method [`String()`](https://pkg.go.dev/fmt?tab=doc#Stringer). Does it make sense to follow the idiomatic naming here?

server/gitlab/gitlab.go
```diff
@@ -47,23 +48,70 @@ type gitlab struct {
...
+	ConfidentialIssuesEvents bool
+	MergeRequestsEvents      bool
+	TagPushEvents            bool
+	NoteEvents               bool
+	JobEvents                bool
+	PipelineEvents           bool
+	WikiPageEvents           bool
+	EnableSSLVerification    bool
+	Token                    string
+}
```
hanzei, line 99: This file begins go get clustered. Let's  move the webhook code to `server/gitlab/webhook.go`.

server/plugin.go
```diff
@@ -335,15 +337,48 @@ func (p *Plugin) sendRefreshEvent(userID string) {
...
+	hasGroupHook, _ := p.HasGroupHook(user, namespace)
+
+	if hasGroupHook {
+		return true, err
+	}
+
+	siteURL := *p.API.GetConfig().ServiceSettings.SiteURL
+	if siteURL == "" {
+		return false, emptySiteURLErr
+	}
```
hanzei, line 358: Instead of checking this everywhere, having the check in `OnActivate()` would allow admins to know right away that this plugin will have issue, if `SiteURL` is not set.

server/plugin.go
```diff
@@ -33,6 +33,8 @@ const (
 	SETTING_OFF           = "off"
 )
 
+var emptySiteURLErr = errors.New("Unable to connect to GitLab")
```
hanzei, line 36: That error message doesn't make clear what a user/admin should do to fix it. Maybe something like:
```suggestion
var emptySiteURLErr = errors.New("SiteURL is not set. Please set it and restart the plugin.")
```

server/command.go
ericjaystevens: creating a webhook  for your CI/CD tooling or another chat or notification service

server/command.go
ericjaystevens: creating a webhook for your CI/CD tooling or another chat or notification service

server/command.go
ericjaystevens: agreed, changed

server/gitlab/api.go
ericjaystevens: My thinking on this one was that it was a formated string and stringify conveyed that.

server/gitlab/gitlab.go
ericjaystevens: Thanks for the nudge, I wanted to do that as well but was hesitant about having two webhook.go files in the same project.

server/command.go
hanzei: Can you elaborate a bit? Is the exact use case that a user wan't to create a webhook to an different service inside of Mattermost?

server/command.go
```diff
@@ -472,15 +473,14 @@ func (p *Plugin) subscribeCommand(parameters []string, channelID string, config
 		}
 	}
 
-	if hasHook {
-		//web hook found
-		return fmt.Sprintf("Successfully subscribed to %s.", fullPath)
+	var hookStatusMessage string
+	if hasHook == false {
```
hanzei, line 477: Nit:
```suggestion
	if !hasHook {
```

server/command.go
ericjaystevens: Yes, An example user story would be a developer creates a new GitLab repo, and wants to get it added to the team's Jenkins server. So the developer messages the Jenkins administrator to get the token and Jenkins URL for the webhook. Then the developer runs the Mattermost Gitlab Plugin command and creates the webhook to Jenkins.

hanzei:
@iomodo Do you have a local GitLab instance setup to test these changes?

iomodo:
@hanzei it's on my todo list.

server/gitlab/gitlab.go
```diff
@@ -47,23 +48,34 @@ type gitlab struct {
...
-
 	if len(g.enterpriseBaseURL) == 0 {
-		return internGitlab.NewOAuthClient(tc, token.AccessToken), nil
+		return internGitlab.NewOAuthClient(token.AccessToken)
 	}
 
-	client := internGitlab.NewOAuthClient(tc, token.AccessToken)
-	if err := client.SetBaseURL(g.enterpriseBaseURL); err != nil {
-		return nil, fmt.Errorf("can't set base url to GitLab client lib: %w", err)
+	client, err := internGitlab.NewOAuthClient(token.AccessToken)
```
iomodo, line 75: Not setting base url will make requests go to the default `gitlab.com` URL.
What was the reason for changing `gitlabConnect` method?

ericjaystevens: You are absolutely right. The reason for the modifications to ```gitlabConnect``` was because of changes in go-gitlab. The types of arguments for ```NewOAuthClient``` changed, and ```SetBaseURL``` became a private function. I must have changed it to get this feature to compile and the forgot that the ```gitlabConnect``` method needs a re-write.

ericjaystevens: I just pushed a change and tested it with Gitlab.com. Here is the link to the breaking change commit that I updated gitlabConnect to address. https://github.com/xanzy/go-gitlab/issues/826

hanzei:
@DHaussermann :point_up:

bbodenmiller:
Looks like https://mattermost.gitbook.io/plugin-gitlab/setup/configuration needs to be updated following this change.

hanzei:
@ericjaystevens Would you be open on updating the documentation?

hanzei:
@bbodenmiller Thanks for the hint :+1:

Issue #175, opened 2020-05-28
Add slash command autocomplete functionality


iomodo:
From 5.24 MM server has a slash command autocomplete functionality feature. An API can be found [here](https://github.com/mattermost/mattermost-server/blob/master/model/command_autocomplete.go#L28)
To implement the feature add `AutocompleteData` struct to the `Command` when registering a slash command.

Similar PRs:
For Todo plugin https://github.com/mattermost/mattermost-plugin-todo/pull/69
For Agenda plugin https://github.com/mattermost/mattermost-plugin-agenda/pull/40

ericjaystevens:
I would like to work on this.

PR #182, opened 2020-06-14
[GH #175] add autocomplete for gitlab slash command 

ericjaystevens:
add autocomplete for gitlab slash command issue# 175

#### Summary
Adds autocomplete support for slash command

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-gitlab/issues/175


server/command.go
```diff
@@ -484,3 +485,74 @@ func (p *Plugin) subscribeCommand(parameters []string, channelID string, config
...
+	gitlabCommand.AddCommand(disconnect)
+
+	todo := model.NewAutocompleteData("todo", "", "Get a list of unread messages and merge requests awaiting your review")
+	gitlabCommand.AddCommand(todo)
+
+	subscribe := model.NewAutocompleteData("subscribe", "[command]", "Available commands: list")
+	subscribeList := model.NewAutocompleteData("list", "", "List current channel subscritpions")
+	subscribe.AddCommand(subscribeList)
+
+	subscribeChannel := model.NewAutocompleteData("subscribe", "owner[/repo] [features]", "Subscribe the current channel to receive notifications from a project")
```
iomodo, line 505: We already have `subscribe` subcommand adding this will make autocomplete as `/gitlab subscribe subscribe..`

server/command.go
```diff
@@ -484,3 +485,74 @@ func (p *Plugin) subscribeCommand(parameters []string, channelID string, config
...
+	gitlabCommand.AddCommand(settings)
+
+	webhook := model.NewAutocompleteData("webhook", "[command]", "Available Commands: list, add")
+	webhookList := model.NewAutocompleteData("list", "owner/[repo]", "List existing project or group webhooks")
+	webhookList.AddTextArgument("Project path: includes user or group name with optional slash project name", "owner[/repo]", "")
+	webhook.AddCommand(webhookList)
+
+	webhookAdd := model.NewAutocompleteData("add", "owner/[repo] [options] [url] [token]", "Add a project or group webhook")
+	webhookAdd.AddTextArgument("Group or Project path: includes user or group name with optional slash project name", "owner[/repo]", "")
+	webhookAdd.AddTextArgument("[Optional] options: comma-delimited list of actions to trigger a webhook, defaults to all with SSL verification", "[* or *noSSL] or [PushEvents,][TagPushEvents,][Comments,][ConfidentialComments,][IssuesEvents,][ConfidentialIssuesEvents,][MergeRequestsEvents,][JobEvents,][PipelineEvents,][WikiPageEvents,][SSLverification]", "")
```
iomodo, line 550: This might be too much text. Can we make it more user readable?

server/command.go
```diff
@@ -484,3 +485,74 @@ func (p *Plugin) subscribeCommand(parameters []string, channelID string, config
...
+	webhookAdd.AddTextArgument("[Optional] url: URL to be triggered triggered. Defaults to this plugins URL", "[url]", "")
+	webhookAdd.AddTextArgument("[Optional] token: Secret for webhook. Defaults to token used in plugin's settings.", "[token]", "")
+	webhook.AddCommand(webhookAdd)
+
+	gitlabCommand.AddCommand(webhook)
+
+	return gitlabCommand
```
iomodo, line 557: Could we add "help" subcommand as well?

server/command.go
hanzei, line 505: Given that this subcommand has the trigger `subscribe` it will add `subscribe` twice. 
Screenshot from 2020-06-15 11-17-17
Screenshot from 2020-06-15 11-17-23

server/command.go
```diff
@@ -484,3 +485,74 @@ func (p *Plugin) subscribeCommand(parameters []string, channelID string, config
...
+	unsubscribe.AddTextArgument("Project path: includes user or group name with optional slash project name", "owner[/repo]", "")
+	gitlabCommand.AddCommand(unsubscribe)
+
+	me := model.NewAutocompleteData("me", "", "Displays the connected GitLab account")
+	gitlabCommand.AddCommand(me)
+
+	settings := model.NewAutocompleteData("settings", "[setting]", "Update your user settings")
+	settingOptions := []model.AutocompleteListItem{{
+		HelpText: "Turn notifications on/off",
+		Hint:     "",
```
hanzei, line 522: Nit: Drop the empty string

server/command.go
```diff
@@ -484,3 +485,74 @@ func (p *Plugin) subscribeCommand(parameters []string, channelID string, config
...
+	todo := model.NewAutocompleteData("todo", "", "Get a list of unread messages and merge requests awaiting your review")
+	gitlabCommand.AddCommand(todo)
+
+	subscribe := model.NewAutocompleteData("subscribe", "[command]", "Available commands: list")
+	subscribeList := model.NewAutocompleteData("list", "", "List current channel subscritpions")
+	subscribe.AddCommand(subscribeList)
+
+	subscribeChannel := model.NewAutocompleteData("subscribe", "owner[/repo] [features]", "Subscribe the current channel to receive notifications from a project")
+	subscribeChannel.AddTextArgument("Project path: includes user or group name with optional slash project name", "owner[/repo]", "")
+	subscribeChannel.AddTextArgument("Features: comma-delimited list of features to subscribe to", "[issues,][merges,][pushes,][issue_comments,][merge_request_comments,][pipeline,][tag,][pull_reviews,][label:<labelName>]", "")
```
hanzei, line 507: It would be awesome if the features could be auto-selected using a list

server/command.go
ericjaystevens: The only other option I see is to say something like:

```See /gitlab help for available options``` 

Is that preferred?

File: server/command.go (hunk 5 above)
Comment (ericjaystevens) [reply server/command.go:507]: Agreed, but I think it's out of the scope of this PR. Generally, we should probably move away from comma delimitated lists or add it as a supported feature of autocomplete.

File: server/command.go (hunk 1 above)
Comment (ericjaystevens) [reply server/command.go:505]: There seems to be something buggy with the UI when two sets of ```AutocompleteData``` have triggers that start with the same word.

File: go.mod [hunk 6]
```diff
@@ -4,15 +4,19 @@ go 1.13
 
 require (
 	github.com/golang/mock v1.4.3
-	github.com/golang/protobuf v1.3.5 // indirect
 	github.com/hashicorp/go-retryablehttp v0.6.5 // indirect
-	github.com/mattermost/mattermost-server/v5 v5.18.0
-	github.com/pkg/errors v0.8.1
-	github.com/stretchr/testify v1.4.0
+	github.com/mattermost/ldap v3.0.4+incompatible // indirect
+	github.com/mattermost/mattermost-server/v5 v5.3.2-0.20200527124843-dea705969ca4
```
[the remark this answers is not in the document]
ericjaystevens: changed

server/command.go
ericjaystevens: I made changes but i think they expose a UI bug.

server/command.go
ericjaystevens: removed

server/command.go
iomodo: Ahh makes sense. Yes there is no way autocomplete can differentiate, so we should not do that. I'd suggest to use `/gitlab subscriptions` to list them. @hanzei thoughts?
This will change the command itself not only the autocomplete part.

ericjaystevens: Two alternative options to consider

1. Leave the commands the way they are and don't provide autocomplete for either of the subscribe parameter sets. This has the advantage of not changing any command names which could break user workflows.

2. Change ```/gitlab subscribe owner[/repo] [features]``` to ```/gitlab subscribe add owner[/repo] [features]``` This seems more consistent with the slashCommand subCommand action parameters convention then ```/gitlab subscriptions``` but is more verbose then @iomodo suggestion.

Option 2 is my current preference, @iomodo @hanzei what do you think?

server/command.go
```diff
@@ -484,3 +485,72 @@ func (p *Plugin) subscribeCommand(parameters []string, channelID string, config
 
 	return fmt.Sprintf("Successfully subscribed to %s.%s", fullPath, hookStatusMessage)
 }
+
+func getAutocompleteData() *model.AutocompleteData {
+	gitlabCommand := model.NewAutocompleteData("gitlab", "[command]", "Avaliable commands: connect, disconnect, todo, subscribe, unsubscribe, me, settings, webhook")
```
aaronrothschild, line 490:
```suggestion
	gitlabCommand := model.NewAutocompleteData("gitlab", "[command]", "Available commands: connect, disconnect, todo, subscribe, unsubscribe, me, settings, webhook")
```

server/command.go
hanzei: I quite like the ideas as it makes the command much clearer.

server/command.go
```diff
@@ -484,3 +485,72 @@ func (p *Plugin) subscribeCommand(parameters []string, channelID string, config
...
+	gitlabCommand.AddCommand(connect)
+
+	disconnect := model.NewAutocompleteData("disconnect", "", "disconnect your GitLab account")
+	gitlabCommand.AddCommand(disconnect)
+
+	todo := model.NewAutocompleteData("todo", "", "Get a list of unread messages and merge requests awaiting your review")
+	gitlabCommand.AddCommand(todo)
+
+	subscribe := model.NewAutocompleteData("subscribe", "[command]", "Available commands: list")
+	subscribeList := model.NewAutocompleteData("list", "", "List current channel subscritpions")
```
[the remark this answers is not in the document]
ericjaystevens: @malteger Thanks for the correction.

server/command.go
ericjaystevens: @aaronrothschild Thanks for the spell-check.

server/command.go
ericjaystevens: Thanks for the feedback everyone, I'll modify the PR to use Option 2. ```/gitlab subscribe add```

server/command.go
ericjaystevens: Testing with it a little more I like presenting all the options the way it is. I'm happy to change it to the help message if you want though. Autocomplete really shines with a complex command like ```webhook add```

server/command.go
```diff
@@ -481,6 +472,103 @@ func (p *Plugin) subscribeCommand(parameters []string, channelID string, config
...
+	connect := model.NewAutocompleteData("connect", "", "Connect your GitLab account")
+	gitlabCommand.AddCommand(connect)
+
+	disconnect := model.NewAutocompleteData("disconnect", "", "disconnect your GitLab account")
+	gitlabCommand.AddCommand(disconnect)
+
+	todo := model.NewAutocompleteData("todo", "", "Get a list of unread messages and merge requests awaiting your review")
+	gitlabCommand.AddCommand(todo)
+
+	subscribe := model.NewAutocompleteData("subscribe", "[command]", "Available commands: Add, list")
```
hanzei, line 517: Maybe `subscriptions` makes more sense here, as `/gitlab subscriptions add` sounds more like a real sentence then `gitlab subscribe add`. Or `subscription`?

ericjaystevens: You've convinced me, ```subscriptions``` better represents that this is a subcommand, not an action. Especially when adding ```/gitlab subscriptions delete```. Might want also want to consider whether we like plural or singular subcommands. Most CLI's prefer singular, but plural might make the commands closer to natural language. 

My only other concern is that the GitLab and GitHub subcommands are consistent.

hanzei: There is an open PR to add autocomplete to the GitHub plugin: https://github.com/mattermost/mattermost-plugin-github/pull/309. The changes here should be applied there.

hanzei: @aaronrothschild @iomodo What do you think about the `/gitlab subscriptions [delete|add|list]` idea?

iomodo: I like the `subscriptions` subcommand. But probably this should be an alias not causing a "breaking" change. I'm 0/5 on this, but we could leave `/gitlab subscribe ..` as a working command and leave it out of the autocomplete.  Do our customers use some kind of automation that we might be breaking otherwise? @aaronrothschild ?

server/command.go
```diff
@@ -481,6 +472,103 @@ func (p *Plugin) subscribeCommand(parameters []string, channelID string, config
...
+	subscribeAdd := model.NewAutocompleteData("add", "owner[/repo] [features]", "Subscribe the current channel to receive notifications from a project")
+	subscribeAdd.AddTextArgument("Project path: includes user or group name with optional slash project name", "owner[/repo]", "")
+	subscribeAdd.AddTextArgument("Features: comma-delimited list of features to subscribe to", "[issues,][merges,][pushes,][issue_comments,][merge_request_comments,][pipeline,][tag,][pull_reviews,][label:<labelName>]", "")
+	subscribe.AddCommand(subscribeAdd)
+
+	gitlabCommand.AddCommand(subscribe)
+
+	unsubscribe := model.NewAutocompleteData("unsubscribe", "owner[/repo]", "Unsubscribe the current channel from a repository")
+	unsubscribe.AddTextArgument("Project path: includes user or group name with optional slash project name", "owner[/repo]", "")
+	gitlabCommand.AddCommand(unsubscribe)
```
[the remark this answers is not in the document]
aaronrothschild: Yes, ideally you 'manage" subscriptions.  I think we're doing this in Jira 3.0 as well @mickmister ?

server/command.go
aaronrothschild: Agreed, let's head in the right direction (Subscriptions) but not break the past (Subscribe).

ericjaystevens:
Seem's like the consensus is to change ```/gitlab unsubscribe``` and ```/gitlab subscribe [subcommand]``` to ```/gitlab subscriptions [deleted | add | list ]``` with aliases for  ```/gitlab unsubscribe``` and ```/gitlab subscribe [subcommand]``` I'll move forward with that goal.

We can generalize this that a convention exists for slash subcommands of  ```/slashcommand noun verb arguments```

server/command.go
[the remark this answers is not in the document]
mickmister: @aaronrothschild Not sure what you mean. The Jira subscriptions involve a modal, where this is using just a slash command, so the same verbs don't really apply. Jira's subcommands are `subscribe edit` and `subscribe list`. Maybe I'm misunderstanding your question.

server/command.go
```diff
@@ -62,11 +62,11 @@ const (
 const (
 	groupNotFoundError   = "404 {message: 404 Group Not Found}"
 	groupNotFoundMessage = "Unable to find GitLab group: "
-)
 
-const (
 	projectNotFoundError   = "404 {message: 404 Project Not Found}"
 	projectNotFoundMessage = "Unable to find project with namespace: "
+
+	invalidSubscribeSubCommand = "Invalid subscribe command. Available commands are 'add' and 'list'"
```
hanzei, line 69: And `delete`?

server/command.go
```diff
@@ -398,39 +398,43 @@ func parseTriggers(triggersCsv string) *gitlab.AddWebhookOptions {
...
-func (p *Plugin) subscribeCommand(parameters []string, channelID string, config *configuration, info *gitlab.GitlabUserInfo) string {
-	features := "merges,issues,tag"
+func (p *Plugin) subscriptionDelete(info *gitlab.GitlabUserInfo, config *configuration, fullPath, channelID string) (string, error) {
+	normalizedPath := normalizePath(fullPath, config.GitlabURL)
+	if deleted, err := p.Unsubscribe(channelID, normalizedPath); err != nil {
+		p.API.LogError("can't unsubscribe channel in command", "err", err.Error())
+		return "Encountered an error trying to unsubscribe. Please try again.", nil
+	} else if !deleted {
+		return "Subscription not found, please check repository name.", nil
+	}
```
hanzei, line 408: Nit pick: At mattermost the patter used is usually
```suggestion
	deleted, err := p.Unsubscribe(channelID, normalizedPath);
	if err != nil {
		p.API.LogError("can't unsubscribe channel in command", "err", err.Error())
		return "Encountered an error trying to unsubscribe. Please try again.", nil
	} 
	
	if !deleted {
		return "Subscription not found, please check repository name.", nil
	}
```

server/command.go
ericjaystevens: updated string literals

PR #220, opened 2020-10-04
Update environment.md

justinegeffen:
Replaced content with updated links and content.


hanzei (approved):
LGTM 


One question: Should we remove the duplicate information in `README.md`?

justinegeffen:
Yep; I'll add that into this PR. Good catch!

Issue #184, opened 2020-06-30
Push event notifications contain broken link to Gitlab user page


n-st:
When commits are pushed to a repo, the plugin generates a message of the format "[Jane Doe](https://gitlab.example.com/Jane Doe) has pushed 1 commit to [janedoe/some-repo](https://gitlab.example.com/janedoe/some-repo)".

The link to the user's page is based on their *display name*, not *username*, and is therefore often invalid (404 on the Gitlab side of things).

n-st:
From a quick look at the code it looks like this stems from a confusion about the Gitlab API:
- The notification text is assembled in [server/webhook/push.go, line 55](https://github.com/mattermost/mattermost-plugin-gitlab/blob/master/server/webhook/push.go#L55),
- based on `senderGitlabUsername`,
- [obtained from `event.UserName` from a `gitlab.PushEvent` argument](https://github.com/mattermost/mattermost-plugin-gitlab/blob/master/server/webhook/push.go#L41),
- which, in turn, is a wrapper around the ["user_name"](https://github.com/xanzy/go-gitlab/blob/master/event_webhook_types.go#L37) attribute in [Gitlab's webhook JSON format](https://docs.gitlab.com/ce/user/project/integrations/webhooks.html#push-events).

The actual username (as it would appear in the URL) is aptly named `UserUsername` in the `go-gitlab` library.

n-st:
It seems that `event.UserName` is also used (perhaps accidentally) in `server/webhook.go` and `server/webhook/tag.go`.

I'd rather not set up a full Go development environment for such a small change, but perhaps this patch will be useful. :)
[0001-Fix-Use-Gitlab-username-instead-of-display-name-to-c.patch.txt](https://github.com/mattermost/mattermost-plugin-gitlab/files/4853654/0001-Fix-Use-Gitlab-username-instead-of-display-name-to-c.patch.txt)

n-thumann:
I´d like to work on it, if you don´t mind 😊

PR #225, opened 2020-10-12
[GH-184]: Use UserUsername in Link, Fixes GitLab id storage

n-thumann:
#### Summary

This PR aims to solve the issue with broken account links in GitLab notifications.
It basically replaces `UserName` with `UserUserName` to use the actual GitLab username instead of the displayed name in the link. While fixing that, it was also required to fix the storage of the GitLab user id. Before this commit, `userID` and `GitlabUserID` have been mixed up, so that `GetUsernameByID` wasn´t returning the correct username.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-gitlab/issues/184

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -208,14 +208,16 @@ func (p *Plugin) ExecuteCommand(_ *plugin.Context, args *model.CommandArgs) (*mo
 		case SettingNotifications:
 			if value {
 				if err := p.storeGitlabToUserIDMapping(info.GitlabUsername, info.UserID); err != nil {
-					p.API.LogError("can't store GitLab user id mapping", "err", err.Error())
+					p.API.LogError("can't store GitLab to user id mapping", "err", err.Error())
 					return p.getCommandResponse(args, "Unknown error please retry or ask to an administrator to look at logs"), nil
 				}
-			} else {
-				if err := p.API.KVDelete(info.GitlabUsername + GitlabUsernameKey); err != nil {
-					p.API.LogError("can't delete GitLab username in kvstore", "err", err.Error())
+				if err := p.storeGitlabIDToUserIDMapping(info.GitlabUsername, info.GitlabUserID); err != nil {
+					p.API.LogError("can't store GitLab to GitLab id mapping", "err", err.Error())
 					return p.getCommandResponse(args, "Unknown error please retry or ask to an administrator to look at logs"), nil
 				}
+			} else if err := p.API.KVDelete(info.GitlabUsername + GitlabUsernameKey); err != nil {
+				p.API.LogError("can't delete GitLab username in kvstore", "err", err.Error())
+				return p.getCommandResponse(args, "Unknown error please retry or ask to an administrator to look at logs"), nil
 			}
 			info.Settings.Notifications = value
 		case SettingReminders:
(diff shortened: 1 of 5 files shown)


n-thumann:
By the way, I changed the notification format to `[n-thumann](link) has pushed 1 commit to [n-thumann/test](link)` (notice the GitLab username in the beginning instead of name) to match the other notifications, although I would personally find it better to use the actual name (so `[Nicolas Thumann](link) has pushed`), but in all notifications to be consistent. Let me know if you agree, I´ll change it then ✌️

server/command.go
```diff
@@ -208,14 +208,16 @@ func (p *Plugin) ExecuteCommand(_ *plugin.Context, args *model.CommandArgs) (*mo
...
 					return p.getCommandResponse(args, "Unknown error please retry or ask to an administrator to look at logs"), nil
 				}
-			} else {
-				if err := p.API.KVDelete(info.GitlabUsername + GitlabUsernameKey); err != nil {
-					p.API.LogError("can't delete GitLab username in kvstore", "err", err.Error())
+				if err := p.storeGitlabIDToUserIDMapping(info.GitlabUsername, info.GitlabUserID); err != nil {
+					p.API.LogError("can't store GitLab to GitLab id mapping", "err", err.Error())
 					return p.getCommandResponse(args, "Unknown error please retry or ask to an administrator to look at logs"), nil
 				}
+			} else if err := p.API.KVDelete(info.GitlabUsername + GitlabUsernameKey); err != nil {
```
levb, line 218: 2/5 wrap this into a `p.deleteXXX` function so it's more intention-revealing and consistent with `store...`

server/webhook/tag.go
```diff
@@ -39,7 +39,7 @@ func (w *webhook) handleDMTag(event *gitlab.TagEvent) ([]*HandleWebhook, error)
 }
 
 func (w *webhook) handleChannelTag(event *gitlab.TagEvent) ([]*HandleWebhook, error) {
-	senderGitlabUsername := event.UserName
+	senderGitlabUsername := w.gitlabRetreiver.GetUsernameByID(event.UserID)
```
levb, line 42: Is this needed because `event.UserUsername` does not come in the message?

n-thumann: Yes, it is :) See [here](https://github.com/xanzy/go-gitlab/blob/master/event_webhook_types.go#L86) or [here](https://docs.gitlab.com/ce/user/project/integrations/webhooks.html#tag-events)

server/command.go
n-thumann: Sure, thanks :)

n-thumann: Done, I also refactored two other similar functions so that it´s more consistent :v:

levb (approved):
LGTM, thx for the improvement and the cleanup!

jfrerich:
@iomodo reminder to review this PR.  Thanks for the contribution, @n-thumann!

iomodo (changes requested):
Thanks a lot @n-thumann, great work!
One minor comment below regarding the error wrapping.

n-thumann:
I appreciate your feedback, @iomodo 👏 I just incorporated your comments. Let me know if they´re okay now :)

iomodo (approved):
LGMT! Thanks @n-thumann

DHaussermann (approved):
Tested and passed
- Subscription posts for push events now have working link
- Post structure and link is consisted with that of other events
- Checked other events to ensure they all seem tp have working links

LGTM!

Thanks @n-thumann for this fix!

PR #241, opened 2021-01-22
support FreeBSD

swills:


hanzei (changes requested):
Thanks for the contribution :+1: 

Could you elaborate a bit why you think this change should go into `master`? The Mattermost Marketplace is the default way to install plugin and it doesn't support FreeBSD any way. Who is the audience for this change?

lindy65:
Hi @swills - I'm helping to review stale PRs and, since we haven't heard from you in a few months, I'll close this PR for now. Please feel free to re-open if you're able to prioritize. Thanks!

//cc @hanzei

PR #244, opened 2021-02-10
Fix allowed namespaces

iomodo:
#### Summary
Fixes the allowed namespaces bug.  Namespaces have format `group/project`. We will allow all namespaces that start with `GitlabGroup` 
#### Ticket Link
  Fixes https://mattermost.atlassian.net/browse/MM-32680


hanzei (approved):
Good catch :+1: LGTM

larkox (approved):

DHaussermann (approved):
Tested and passed
- Confirmed DMs work as expected when the scope is restricted to a Group such as `dhaussermann`
- No need to include a project (as this config is invalid for subscriptions)
- Group restriction config is now consistent for DM and subscriptions
No issues found.

LGTM!

PR #260, opened 2021-09-14
fix the link to get the pending reviews

cpanato:
#### Summary
Using the gitlab in our community deployment notice that it did not show any MR that I need to review, checking closely notice that we were using a wrong filter. this PR update the filter and using that I was able to see my pending MR to review

#### Ticket Link
n/a


iomodo (approved):
Good catch! Thanks @cpanato

hanzei (approved):
Good catch :+1:

hanzei:
Skipping QA review as this is a minor fix

PR #284, opened 2022-04-03
[issue 237] Number of Issues, Notes etc. not correct: API call responds are limited #237

### LINKED ISSUE #237: Number of Issues, Notes etc. not correct: API call responds are limited
The plugin shows in Mattermost only 20 issues, notes etc.

In my case. I have 66 Issues and 24 notes in gitlab.
gitlab

But unfortunately the Plugin shows only 20 at maximum.
mm

Same to the note list with `/gitlab todo` list

The problem is, that every API call is limited to 20 respond objects in default and 100 in max.

https://docs.gitlab.com/ee/api/#pagination

Could you please implement an iteration to respect all pagination sites, so that the value for issues, notes etc. are correct?

Thx a lot :)

sibasankarnayak:
implemented pagination for information fetched from gitlab.

ticket here
Fixes #237 

Please find the fix screenshot here
image


server/gitlab/api.go
```diff
@@ -258,24 +298,50 @@ func (g *gitlab) GetYourPrs(user *UserInfo) ([]*internGitlab.MergeRequest, error
 	opened := stateOpened
 	scope := scopeAll
 
-	var result []*internGitlab.MergeRequest
+	var prList []*internGitlab.MergeRequest
 	var errRequest error
+	var resp *internGitlab.Response
+	var result []*internGitlab.MergeRequest
```
mickmister, line 304: `result` is usually used for the "final result", but it's being used as an intermediate value below. Maybe use `current` and `result` to convey the meaning of each variable

server/gitlab/api.go
```diff
@@ -287,24 +353,50 @@ func (g *gitlab) GetYourAssignments(user *UserInfo) ([]*internGitlab.Issue, erro
 	opened := stateOpened
 	scope := scopeAll
 
+	var issues []*internGitlab.Issue
+	var resp *internGitlab.Response
 	var result []*internGitlab.Issue
 	var errRequest error
```
mickmister, line 359: Same comment here. Prefer `current` and `result`

server/gitlab/api.go
```diff
@@ -313,12 +405,25 @@ func (g *gitlab) GetUnreads(user *UserInfo) ([]*internGitlab.Todo, error) {
 		return nil, err
 	}
 
-	result, _, err := client.Todos.ListTodos(&internGitlab.ListTodosOptions{})
-	if err != nil {
-		return nil, errors.Wrap(err, "can't list todo in GitLab api")
+	var results []*internGitlab.Todo
+	var result []*internGitlab.Todo
+	var resp *internGitlab.Response
```
mickmister, line 410: `result` and `current` here as well

server/gitlab/api.go
```diff
@@ -313,12 +405,25 @@ func (g *gitlab) GetUnreads(user *UserInfo) ([]*internGitlab.Todo, error) {
...
+	opt := &internGitlab.ListTodosOptions{
+		ListOptions: internGitlab.ListOptions{Page: 1, PerPage: 1},
+	}
+	for {
+		result, resp, err = client.Todos.ListTodos(opt)
```
mickmister, line 415: I think we only need to declare `result` in the upper scope if we write it like this:

```suggestion
		current, resp, err := client.Todos.ListTodos(opt)
```

Notice the colon by the equals sign

mickmister (changes requested):

server/gitlab/api.go
```diff
@@ -198,10 +198,25 @@ func (g *gitlab) GetProjectHooks(user *UserInfo, owner string, repo string) ([]*
...
+	opt := &internGitlab.ListProjectHooksOptions{
+		PerPage: 20,
+		Page:    1,
+	}
+
+	var projectHooks []*internGitlab.ProjectHook
+	var hooks []*internGitlab.ProjectHook
+	var resp *internGitlab.Response
+	for {
+		projectPath := fmt.Sprintf("%s/%s", owner, repo)
```
hanzei, line 210: Nit pick: Does it need to be recalculated each iteration?

server/gitlab/api.go
```diff
@@ -198,10 +198,25 @@ func (g *gitlab) GetProjectHooks(user *UserInfo, owner string, repo string) ([]*
 		return nil, err
 	}
 
-	projectPath := fmt.Sprintf("%s/%s", owner, repo)
-	projectHooks, _, err := client.Projects.ListProjectHooks(projectPath, nil)
-	if err != nil {
-		return nil, err
+	opt := &internGitlab.ListProjectHooksOptions{
+		PerPage: 20,
```
hanzei, line 202: Let's make this a const

server/gitlab/api.go
```diff
@@ -198,10 +198,25 @@ func (g *gitlab) GetProjectHooks(user *UserInfo, owner string, repo string) ([]*
...
+		Page:    1,
```
hanzei, line 203: Page 1 or 0?

hanzei (changes requested):

server/gitlab/api.go
sibasankarnayak: @hanzei it starts from 1 can refer here https://docs.gitlab.com/ee/api/index.html#pagination
image

hanzei (approved):
LGTM

mickmister (approved):

hanzei:
@DHaussermann Gentle reminder to review the PR

DHaussermann (approved):
Tested and passed
When the records count exceeds 20 - accurate counts are now reflected.
No regressions found.
LGTM!

hanzei:
Replaced by https://github.com/mattermost/mattermost-plugin-gitlab/pull/366

Issue #61, opened 2019-06-05
Backport github plugin improvements


manland:
- [ ] https://github.com/mattermost/mattermost-plugin-github/pull/103
- [ ] https://github.com/mattermost/mattermost-plugin-github/pull/170
- [ ] https://github.com/mattermost/mattermost-plugin-github/pull/123
- [ ] https://github.com/mattermost/mattermost-plugin-github/pull/137
- [ ] https://github.com/mattermost/mattermost-plugin-github/pull/258

hanzei:
@hanzei to update the list

PR #299, opened 2022-06-02
[GH-61]: Show code block previews of Gitlab permalinks

shivamjosh:
#### Summary
For every message, we now hook into the MessageWillBePosted event and check
if it contains a permalink to a Gitlab file.

If yes, we parse out the link and make an API call to the Gitlab files API.
Get the file contents and filter out the required lines given in the permalink anchor.

Then we construct a code block in markdown and replace the original string.

Added a config key to gate this feature.

Screenshots:

With config disabled:
image

After enabling it:
image

The newly formed hyperlink is a link to the original permalink

Config setting:
image

#### Ticket Link

  Fixes https://github.com/mattermost/mattermost-plugin-gitlab/issues/61


server/plugin.go
```diff
@@ -72,6 +73,15 @@ type Plugin struct {
 
 	WebhookHandler webhook.Webhook
 	GitlabClient   gitlab.Gitlab
+	// gitlabPermalinkRegex is used to parse gitlab permalinks in post messages.
+	gitlabPermalinkRegex *regexp.Regexp
+}
+
+// NewPlugin returns an instance of a Plugin.
+func NewPlugin() *Plugin {
+	return &Plugin{
```
levb, line 82: 0/5 a static `var gitlabPermalinkRegex = regexp.Must...` would be a touch more idiomatic, since it never changes and does not need to be per-plugin?

server/plugin.go
```diff
@@ -146,6 +156,32 @@ func (p *Plugin) OnPluginClusterEvent(c *plugin.Context, ev model.PluginClusterE
 	p.HandleClusterEvent(ev)
 }
 
+func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*model.Post, string) {
```
levb, line 159: 0/5 It would appear that in hi-traffic environments we should be using `MessageHasBeenPosted`, which is executing asynchronously and not `MessageWillBePosted` which is executed in-line.

I am very uneasy about tying up an RPC "connection" (state) to make other RPC, let alone HTTP requests. cc @agnivade for another opinion, it may be that the overhead here is not as significant as I think.

server/permalinks.go
```diff
@@ -0,0 +1,151 @@
...
+// permalinkLineContext shows the number of lines before and after to show
+// if the link points to a single line.
+const permalinkLineContext = 3
+
+// replacement holds necessary info to replace GitLab permalinks
+// in messages with a code preview block.
+type replacement struct {
+	index         int      // index of the permalink in the string
+	word          string   // the permalink
+	permalinkInfo struct { // holds the necessary metadata of a permalink
```
levb, line 33: 0/5 define a `type permalinkInfo struct` as ^^ and use it in the test, for compactness.

server/permalinks.go
```diff
@@ -0,0 +1,151 @@
...
+		user   string
+		repo   string
+		path   string
+		line   string
+	}
+}
+
+// getReplacements returns the permalink replacements that need to be performed
+// on a message. The returned slice is sorted by the index in ascending order.
+func (p *Plugin) getReplacements(msg string) []replacement {
```
levb, line 45: 1/5 s/getReplacements/getPermalinkReplacements/

server/permalinks.go
```diff
@@ -0,0 +1,151 @@
...
+			p.API.LogError("bad git commit hash in permalink", "error", err.Error(), "hash", r.permalinkInfo.commit)
+			continue
+		}
+
+		// get the file contents
+		opts := gitlab.GetFileOptions{
+			Ref: &r.permalinkInfo.commit,
+		}
+		projectPath := fmt.Sprintf("%s/%s", r.permalinkInfo.user, r.permalinkInfo.repo)
+		// TODO: make all of these requests concurrently.
```
levb, line 107: 0/5 Can we please do this now? Especially in the context of my `MessageWillBePosted` remark.

levb (changes requested):
Mostly LGTM, with nits.

The big question I have if we should be using `MessageHasBeenPosted` here instead, to rewrite the post asynchronously.

server/plugin.go
agnivade: `MessageWillBePosted` is necessary if a plugin needs to modify the incoming post, which in this case is needed.

server/utils.go
```diff
@@ -141,3 +144,119 @@ func isValidURL(rawURL string) error {
...
+//
+// It is a poor man's version of checking markdown hyperlinks without
+// using a full-blown markdown parser. The idea is to quickly confirm
+// whether a permalink is inside a markdown link or not. Something like
+// "text ]( permalink" is rare enough. Even then, it is okay if
+// there are false positives, but there cannot be any false negatives.
+//
+// Note: it is fine to go one byte at a time instead of one rune because
+// we are anyways looking for ASCII chars.
+func isInsideLink(msg string, index int) bool {
```
mickmister, line 222: Is this implementation or explanation from another source?

server/permalinks.go
```diff
@@ -0,0 +1,151 @@
...
+
+// makeReplacements performs the given replacements on the msg and returns
+// the new msg. The replacements slice needs to be sorted by the index in ascending order.
+func (p *Plugin) makeReplacements(msg string, replacements []replacement, glClient *gitlab.Client) string {
+	// iterating the slice in reverse to preserve the replacement indices.
+	for i := len(replacements) - 1; i >= 0; i-- {
+		r := replacements[i]
+		// quick bailout if the commit hash is not proper.
+		if _, err := hex.DecodeString(r.permalinkInfo.commit); err != nil {
+			p.API.LogError("bad git commit hash in permalink", "error", err.Error(), "hash", r.permalinkInfo.commit)
```
mickmister, line 98: I'm thinking this should be a debug log, since this isn't a critical issue. It's just a malformed URL in a post body.

server/permalinks.go
```diff
@@ -0,0 +1,151 @@
...
+			}
+		}
+		replacements = append(replacements, r)
+	}
+	return replacements
+}
+
+// makeReplacements performs the given replacements on the msg and returns
+// the new msg. The replacements slice needs to be sorted by the index in ascending order.
+func (p *Plugin) makeReplacements(msg string, replacements []replacement, glClient *gitlab.Client) string {
```
mickmister, line 92: Was this code adapted from the GitHub plugin's implementation of this feature?

server/permalinks.go
```diff
@@ -0,0 +1,151 @@
...
+		opts := gitlab.GetFileOptions{
+			Ref: &r.permalinkInfo.commit,
+		}
+		projectPath := fmt.Sprintf("%s/%s", r.permalinkInfo.user, r.permalinkInfo.repo)
+		// TODO: make all of these requests concurrently.
+		_, cancel := context.WithTimeout(context.Background(), permalinkReqTimeout)
+		file, _, err := glClient.RepositoryFiles.GetFile(projectPath, r.permalinkInfo.path, &opts)
+		defer cancel()
+		if err != nil {
+			p.API.LogError("error while fetching file contents", "error", err.Error(), "path", r.permalinkInfo.path)
```
mickmister, line 112: I'm thinking every log message in this function should be a debug log

server/permalinks_test.go
```diff
@@ -0,0 +1,420 @@
...
+}
+
+const (
+	baseURLPath = "/api/v4"
+)
+
+func getClient() (*gitlab.Client, func()) {
+	apiHandler := http.NewServeMux()
+	apiHandler.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) {
+		if req.URL.Path == "/api/v4/projects/mattermost/mattermost-server/repository/files/app/authentication.go" {
```
mickmister, line 399: Shouldn't we use the constant on the last line here?

server/permalinks_test.go
```diff
@@ -0,0 +1,420 @@
...
+			msg := p.makeReplacements(tc.input, tc.replacements, client)
+			assert.Equalf(t, tc.output, msg, "mismatched output")
+		})
+	}
+
+	mockPluginAPI.AssertCalled(t, "LogError", "bad git commit hash in permalink", "error", "encoding/hex: invalid byte: U+0068 'h'", "hash", "badhash")
+}
+
+const (
+	baseURLPath = "/api/v4"
```
mickmister, line 393: Since we're declaring this in the middle of the file, and only declaring one constant:

```suggestion
const baseURLPath = "/api/v4"
```

server/plugin.go
```diff
@@ -146,6 +156,32 @@ func (p *Plugin) OnPluginClusterEvent(c *plugin.Context, ev model.PluginClusterE
...
+	if post.UserId == "" {
+		return nil, ""
+	}
+
+	msg := post.Message
+	info, err := p.getGitlabUserInfoByMattermostID(post.UserId)
+	if err != nil {
+		p.API.LogError("error in getting user info", "error", err.Message)
+		return nil, ""
+	}
```
mickmister, line 174: What if the user simply isn't connected to GitLab? We don't want to log this as an "error" in this case. Also, we'll want to be more specific with these error messages, saying that we are processing a post's permalinks, to differentiate from an intentional action related to GitLab.

I'm thinking all `LogError`'s in this PR should be changed to `LogDebug` due to the passive nature of this feature

mickmister (changes requested):
Thanks for this @shivamjosh :+1:

I have a few requests including one regarding performance, and a few questions.

I also have a request that when there has been prior review on the work, to please link to the existing PR in the new PR's description: https://github.com/Brightscout/mattermost-plugin-gitlab/pull/4

GitHub isn't allowing me to reply on conversations there so I'm copying this discussion here:

https://github.com/Brightscout/mattermost-plugin-gitlab/pull/4#discussion_r885233011

> Have you checked if there's a difference between Github and Gitlab permalinks or just copied this regex from Github plugin?

> I have copied it from Github plugin and then made changes according to the GitLab plugin.

How confident are we that the URL formats between GitHub and GitLab are generally identical?

server/permalinks.go
raghavaggarwal2308: Yes

server/utils.go
raghavaggarwal2308: Yes, this implementation is taken from the Github plugin.

server/plugin.go
raghavaggarwal2308: Done

server/permalinks.go
raghavaggarwal2308: Done

server/permalinks.go
raghavaggarwal2308: Done

server/permalinks.go
raghavaggarwal2308: Done

server/permalinks.go
raghavaggarwal2308: Done

server/permalinks.go
raghavaggarwal2308: Done

server/permalinks_test.go
raghavaggarwal2308: Done

server/permalinks_test.go
raghavaggarwal2308: Now, we have more than one constant i.e `baseURLPath` and `reqURLPath`

server/plugin.go
raghavaggarwal2308: Done

raghavaggarwal2308:
@mickmister @levb Fixed the review fixes you mentioned.

server/permalinks.go
```diff
@@ -0,0 +1,169 @@
...
+			case "line":
+				r.permalinkData.line = m[j]
+			}
+		}
+		replacements = append(replacements, r)
+	}
+	return replacements
+}
+
+func (p *Plugin) processReplacement(r replacement, glClient *gitlab.Client, wg *sync.WaitGroup, markdownForPermalink []string, index int) {
```
levb, line 93: 0/5 maybe worth a quick comment that concurrently modifying unique slice elements is not racy.

levb (commented):
LGTM (upon resolving @mickmister's comments)

server/permalinks.go
raghavaggarwal2308: Done

server/utils.go
mickmister: Can we link to the code here using a GitHub permalink?

server/permalinks.go
mickmister: Can we put a comment linking to the original function using a GitHub permalink?

mickmister (approved):
LGTM, though I'd like this pending question to be answered:

Discussion:

> Have you checked if there's a difference between Github and Gitlab permalinks or just copied this regex from Github plugin?

> I have copied it from Github plugin and then made changes according to the GitLab plugin.

Question:

How confident are we that the URL formats between GitHub and GitLab are generally identical?

raghavaggarwal2308:
@mickmister In GitLab, we have two types of permalinks - 
1. When we go to a file and copy the permalink directly
    Example: https://gitlab.com/raghavaggarwal2308/Calculator/-/blob/test/index.html#L7
    - This permalink contains the branch name in the URL.
  
2. When we go to the file and click on the `Permalink` button on the top right and then copy the permalink

Screenshot from 2022-08-19 23-47-29


    Example: https://gitlab.com/raghavaggarwal2308/Calculator/-/blob/0d2465ed07d4b314ec0f9a838bb2f064e34dc0d7/index.html#L7
    - This permalink contains the commit ID in the URL
    
    But in Github, we have only one type of permalink i.e: permalink with the commit ID
    Example: https://github.com/raghavaggarwal2308/Hamagi/blob/7322e6554c839af1da0bfb30d71b070f473111bf/shopping/src/App.js#L13
    
    Our current code is working for the permalink containing the commit ID and not for the permalink with the branch name.

catalintomai (approved):

raghavaggarwal2308:
We are working on fixing the above issue and will submit it by tomorrow.

server/permalinks.go
raghavaggarwal2308: Done

server/utils.go
raghavaggarwal2308: Done

raghavaggarwal2308:
@mickmister Fixed the issue and added the comments to link the code with Github plugin.

raghavaggarwal2308:
@levb @mickmister Gentle reminder for re-review

raghavaggarwal2308:
@levb @mickmister Gentle reminder for re-review

raghavaggarwal2308:
@levb @mickmister Gentle reminder for re-review

raghavaggarwal2308:
@levb @mickmister Gentle reminder for re-review

raghavaggarwal2308:
@mickmister @levb Gentle reminder for re-review

levb (approved):

mickmister:
> Our current code is working for the permalink containing the commit ID and not for the permalink with the branch name.

What's the complexity of supporting both?

mickmister:
@raghavaggarwal2308 Sorry I meant to mention you in the above comment

mickmister (approved):

raghavaggarwal2308:
@mickmister We are supporting both not

raghavaggarwal2308:
@mickmister Gentle reminder for re-review

raghavaggarwal2308:
@mickmister gentle reminder for re-review

mickmister:
@raghavaggarwal2308 There are no changes from when I approved the PR

raghavaggarwal2308:
@mickmister Apologies for this. Can you please remove the `Dev-review` label if all the dev reviews are complete

mickmister:
> > Our current code is working for the permalink containing the commit ID and not for the permalink with the branch name.
> 
> What's the complexity of supporting both?

> @mickmister We are supporting both now

@raghavaggarwal2308 Does this mean this was committed before my comment? Just want to make sure I understand what's going on

raghavaggarwal2308:
> > > Our current code is working for the permalink containing the commit ID and not for the permalink with the branch name.
> > 
> > 
> > What's the complexity of supporting both?
> 
> > @mickmister We are supporting both now
> 
> @raghavaggarwal2308 Does this mean this was committed before my comment? Just want to make sure I understand what's going on

@mickmister Yes, it was committed before that. Mentioned it in [this](https://github.com/mattermost/mattermost-plugin-gitlab/pull/299#issuecomment-1225436381) comment.

raghavaggarwal2308:
@DHaussermann Gentle reminder for review

hanzei:
@DHaussermann Gentle reminder to review the PR

DHaussermann:
Hi @shivamjosh 

I'm not seeing this work as expected. I see only the normal link preview regardless of if the new code block feature is enabled. 

Maybe the GitLab functionality is different from what I'm trying? With this link `https://gitlab.com/dhaussermann/dkh-public/-/blame/323cdf25b439e9d01bcb139859b39fc4580e65a1/scipt.js#L11-L12` should I be seeing a code block? Your example in the summaty only points to 1 line but, I have tried this as well.   

image

image

Please let me know if you can provide some detail on this.

raghavaggarwal2308:
@DHaussermann You are using the "blame" link. Please use a normal permalink (Ex: https://gitlab.com/raghavaggarwal2308/Calculator/-/blob/main/app.js#L12)
I think "blame" links are not supported on Github plugin as well

On github plugin:
image

DHaussermann:
@raghavaggarwal2308 Thanks for the guidance on this. I was struggling a bit with the GitLab UI to get a link to more than 1 line of code. I did not realize I was getting the `blame` link instead of the `blob` link. 🤦‍♂️ 

That said - I have now done some testing and the feature is woking. 
I do have 1 follow-up point to ask about. Regarding private repos it seems we don't fetch the code block previews. Even if private repos is enabled in plugin config AND the user making the post is connected in GitLab as a permission user - The code block does not get retrieved.

Can you confirm if this is the expected behavior?

raghavaggarwal2308:
@DHaussermann This feature should work for private repos as well. It's working fine on my machine. Can you share some details? Like the link you are using or the server logs when you posted the link in the channel.

DHaussermann:
@raghavaggarwal2308 When I use the. same steps to make a code link `https://gitlab.com/dkh-test-group/dkh-project2/-/blob/master/Text.txt#L2-L4` I see this posting a preview now. Sorry I must have confused the test users. 

However - Should it be posting code previews if Private repos is disabled as I'm seeing?  Or is that flag not related to code preview?

raghavaggarwal2308:
@DHaussermann If we take reference from the GitHub plugin. This feature is independent of the "Enable Private Repositories" setting. But there is a separate setting for code previews i.e "Enable Code Previews"(see the screenshot attached).
Do you think we should add a similar one in GitLab as well?

DHaussermann:
@raghavaggarwal2308 Yes. We would need some type of toggle on this where the plugin is explicitly configured to allow code block from a private repo.

If this is feature is not ware of the private repo flag - implementing a. drop down would be the ideal solution.

raghavaggarwal2308:
@DHaussermann Added the dropdown in the config setting
Screenshot from 2023-04-17 14-57-02

DHaussermann:
Sorry @raghavaggarwal2308 I missed that this was ready due to the label.

I'll look shortly.

DHaussermann:
@raghavaggarwal2308 I have retested this and this is working as expected. Thank you 👍 

I have only 1 small ask. Now that there is a separate piece of functionality involving private repos, the exiting filed no longer affects anything related to private repos.
image

For this PR - Can you please add some text to the description of the existing filed that will clarify the scope.   
`(Optional) Allow the plugin to work with private repositories.` could be changed to `(Optional) Allow the plugin to work with private repositories for subscriptions.`

This will add context that the code previews are not covered by the existing switch.

raghavaggarwal2308:
@DHaussermann Update the description of EnablePrivateRepos

hanzei:
@DHaussermann Heads up that the PR is ready for your review.

DHaussermann (approved):
Tested and passed

The last point has now been addressed. The label on the exiting radio button has been updated to clarify the scope. 

LGTM!

@raghavaggarwal2308  thanks for the effort on this. Sorry it took me so long to circle back.

DHaussermann:
@hanzei I had applied the `Do Not Merge` label for an unrelated reason. We still have not cut a GitLab release for `v1.7.0` which has the oAuth token fix. 

I am reluctant to merge new features into `master` until we get this cut. Do you have any thoughts on this? I'll start a thread in the channel on Community.

DHaussermann:
@hanzei I'm okay with including this into `1.7.0` is we can merge is soon and it doesn't add significant delays.

PR #301, opened 2022-06-06
[GH-61]: Add tooltip for pull request and issue links

shivamjosh:
#### Summary
This PR adds a tooltip for the pull requests and issue links.

Screenshots:
Issue tooltip:
image

Merge request tooltip:
image

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-gitlab/issues/61

PR #306, opened 2022-06-20
Adding create issue modal

### LINKED ISSUE #61: Backport github plugin improvements

raghavaggarwal2308:
### Summary

Added a modal to create a gitlab issue on an existing project of the user.

- Created API to get all projects of user, project labels, project assignees, project milestones and create a issue.
- Created a "CreateIssueModal" component to create a issue on gitlab.
- Created a "CreateIssuePostMenuAction" component and registered it as "PostDropdownMenuComponent" to open the create issue modal.
- Configured babel-loader to load .tsx and .ts files.
- Upgraded the dependencies in package.json to match the dependencies in github plugin.

### Screenshots

- Click "Create Gitlab Issue" in context menu (shown when user connects his/her gitlab account with plugin installed) on non system messages
Screenshot from 2022-06-20 23-18-40

- Create Gilab issue modal will open
Screenshot from 2022-06-20 23-21-30

- Search and select a Project and issue title
Screenshot from 2022-06-20 23-22-54

- Select labels, assigness, milestone and description if any and click submit to create issue
Screenshot from 2022-06-20 23-23-54

### Ticket
Fixes #61 (Adding create issue modal)


catalintomai (approved):

raghavaggarwal2308:
@mickmister Gentle reminder for review

raghavaggarwal2308:
@mickmister Gentle reminder for review

server/command.go
```diff
@@ -314,6 +320,23 @@ func (p *Plugin) handleSetup(c *plugin.Context, args *model.CommandArgs, paramet
...
+
+	command := parameters[0]
+	parameters = parameters[1:]
+
+	switch {
+	case command == "create":
+		p.openIssueCreateModal(args.UserId, args.ChannelId, strings.Join(parameters, " "))
+		return ""
+	default:
+		return fmt.Sprintf("Unknown subcommand %v", command)
```
spirosoik, line 336: I think the message would be better if we have something like `This command is not implemented yet`

server/plugin.go
```diff
@@ -599,6 +611,30 @@ func (p *Plugin) HasGroupHook(ctx context.Context, user *gitlab.UserInfo, namesp
 	return found, err
 }
 
+// getUsername returns the GitLab username for a given Mattermost user,
+// if the user is connected to GitLab via this plugin.
+// Otherwise it returns the Mattermost username.
+func (p *Plugin) getUsername(userID string) (string, *APIErrorResponse) {
+	info, err := p.getGitlabUserInfoByMattermostID(userID)
+	if err != nil {
```
spirosoik, line 619: minor: I think we can make this a bit more readable and it's ok to have some repetition.
```go
if  err != nil && if err.ID != APIErrorIDNotConnected {
      return "", err
}

if err != nil {
 // rest of logic
}

**Review by spirosoik — COMMENTED** [review]
@raghavaggarwal2308 thanks a lot for your contribution. Overall LGTM, a few minor comments for discussion.

File: server/api.go [hunk 3]
```diff
@@ -557,6 +566,211 @@ func (p *Plugin) getYourAssignments(c *UserContext, w http.ResponseWriter, r *ht
...
+	}
+
+	reply := &model.Post{
+		Message:   message,
+		ChannelId: channelID,
+		RootId:    rootID,
+		UserId:    p.BotUserID,
+	}
+
+	if post != nil {
```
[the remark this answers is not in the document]
raghavaggarwal2308: This behavior was taken from the GitHub plugin. But if you want us to change this when we can do that?

raghavaggarwal2308:
@spirosoik FIxed the review fixes you mentioned

server/api.go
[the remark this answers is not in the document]
spirosoik: @mickmister thoughts around it?

spirosoik (approved):
LGTM

raghavaggarwal2308:
@mickmister Gentle reminder for re-review

raghavaggarwal2308:
@mickmister Gentle reminder for review

server/api.go
```diff
@@ -557,6 +566,211 @@ func (p *Plugin) getYourAssignments(c *UserContext, w http.ResponseWriter, r *ht
 	p.writeAPIResponse(w, result)
 }
 
+func (p *Plugin) createIssue(c *UserContext, w http.ResponseWriter, r *http.Request) {
+	var issue *gitlab.IssueRequest
+
+	if err := json.NewDecoder(r.Body).Decode(&issue); err != nil {
+		c.Log.WithError(err).Warnf("error decoding the JSON body to create issue")
+		p.writeAPIError(w, &APIErrorResponse{ID: "", Message: fmt.Sprintf("please provide a valid JSON object. Error: %s", err.Error()), StatusCode: http.StatusBadRequest})
```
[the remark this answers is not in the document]
mickmister: Yeah maybe we instead say something like "There was an error creating the issue". This error shouldn't ever happen in the case of real use from the frontend.

mickmister (commented):
@raghavaggarwal2308 Sorry I will try to get to this soon. There are some mostly neglible comments I have pending I'll submit now

raghavaggarwal2308:
@mickmister Fixed the review comments you mentioned

webapp/src/components/gitlab_issue_selector.tsx
```diff
@@ -0,0 +1,149 @@
...
+                return [];
+            }
+
+            return issues.map((issue) => {
+                const projectParts = issue.web_url.split('/');
+                let prefix = '';
+                // Extract "username/projectName" from the issueURL parts
+                if (projectParts.length >= 5) {
+                    prefix = `${projectParts[projectParts.length - 5]}/${projectParts[projectParts.length - 4]}`;
+                }
```
mickmister, line 79: Is it possible that the URL doesn't match what want to to here? Maybe there are somtimes other suffixes on the URL that make this `length - 4` etc. thing not work properly?

server/gitlab/api.go
```diff
@@ -383,6 +398,163 @@ func (g *gitlab) GetUnreads(ctx context.Context, user *UserInfo) ([]*internGitla
...
+		return nil, respErr
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	return result, nil
+}
+
+func (g *gitlab) GetAssignees(ctx context.Context, user *UserInfo, projectID string) ([]*internGitlab.ProjectMember, error) {
```
mickmister, line 464: Should this be called `GetProjectMembers`?

mickmister (approved):
No blocking concerns found. Nice work @raghavaggarwal2308 :+1:

DHaussermann:
Hi @raghavaggarwal2308,
I've done a round of testing on this and it's looking pretty good! 👍  I have a bit more work to do but, I wanted to raise some points with you:
1. It looks like the create option always has the selected version of the light theme applied even when the option does not have focus and a different theme is applied.
image
image
2. The Issue Title box accepts 1 too many characters so it can exceed the GitLab max. This is an edge case but, can we set the field to have a max 1 character shorter so that it is impossible to try creating an issue with a title too long for the GitLab side? 
image
3. Also the error is cryptic ☝️ . If there are similar ones - I wonder if you could trim the special characters out`{}[]`      
4. The business logic for which issues/projects you can select for for attach is a bit odd. I am aware that this is also the case for GitHub plugin but, I'm not sure we should copy it. Create has business logic to show only the users projects.(I'm not sure exactly what it is) But Attach will allow the active user to search for and select any project where issues are configured that any GitLab user can add a comment.

image

image

Can we consider just allowing attach to work with the same projects as create rather than matching the GitHub logic here?
cc @asatkinson who may also have thoughts on this point.

webapp/src/components/gitlab_issue_selector.tsx
raghavaggarwal2308: Now we are traversing the array from starting index

raghavaggarwal2308:
@mickmister Fixed the review comments you mentioned

@DHaussermann Working on the comments you mentioned.

webapp/src/components/gitlab_issue_selector.tsx
mickmister: Can you write a unit test for this functionality?

Nityanand13: @mickmister I did the setup to write unit tests in [PR: #301](https://github.com/mattermost/mattermost-plugin-gitlab/pull/301). So should I wait for that PR to be merged or do setup here as well and write unit tests?

mickmister: @Nityanand13 I suppose you'll need to port the setup here as well, as I'm not sure when the other PR will be merged

raghavaggarwal2308:
@DHaussermann Should we make the changes suggested in point 4 or should we wait for the reply of @asatkinson.

raghavaggarwal2308:
> @DHaussermann Should we make the changes suggested in point 4 or should we wait for the reply of @asatkinson.
@DHaussermann Any updates on this?

DHaussermann:
@raghavaggarwal2308 yes your suggestion above is an option. This is how GitHub works. I just find it very awkward to use when there is no Org lock in place so, I would consider that we do not need to match that business logic.

What is the business logic for the repos in the create issue drop down?

DHaussermann:
> @DHaussermann In create issue drop down we get all the projects of the connected user.

hmmm... I see. Yes this does not feel appropriate here for attach. Maybe the best thing to do is keep as is AND make sure it also works like GitHub in that if there is an Org lock in place it filters the repos. 

@asatkinson do you have any objections to this? We already go it for GitHub it just looks very sloppy when Org lock is off.

mickmister (approved):

mickmister (approved):

mickmister:
@DHaussermann This is ready for review

raghavaggarwal2308:
@DHaussermann Gentle reminder for re-review

raghavaggarwal2308:
@DHaussermann Gentle reminder for re-review

raghavaggarwal2308:
@DHaussermann Gentle reminder for re-review

raghavaggarwal2308:
@DHaussermann Gentle reminder for re-review

server/gitlab/api.go
```diff
@@ -681,8 +681,8 @@ func (g *gitlab) GetYourProjects(ctx context.Context, user *UserInfo) ([]*intern
 	return projects, nil
 }
 
-func (g *gitlab) GetLabels(ctx context.Context, user *UserInfo, projectID string) ([]*internGitlab.Label, error) {
-	client, err := g.gitlabConnect(*user.Token)
+func (g *gitlab) GetLabels(ctx context.Context, user *UserInfo, projectID string, token *oauth2.Token) ([]*internGitlab.Label, error) {
+	client, err := g.gitlabConnect(*token)
```
mickmister, line 685: Do we need to use `useGitlabClient` here?

raghavaggarwal2308: @mickmister We can use it but the main use of `useGitlabClient` is to fetch the oauth token, but here we already have the token in the params. Should we still use it here?

mickmister: Oh right, this `gitlab` struct doesn't use `useGitlabClient` function

raghavaggarwal2308:
@DHaussermann Gentle reminder for re-review

AayushChaudhary0001:
While testing this PR, I faced an issue that can be reproduced by following the steps below.

**Steps to reproduce:**

1. Click on the Create GitLab Issue button.
2. Select the desired project in the modal.
3. Multiple fields might appear in the modal after selecting the project from the dropdown.
4. Delete the selected project.
5. Try to open the additional fields added in the modal itself.

**Expected result:-**
Since the project has been removed from the project dropdown, no further data should be fetched in the extra fields added in the modal associated with the project.

**Actual result:-**
The extra dropdowns added can fetch the data (labels, assignees) of the project without even selecting the project.

image

raghavaggarwal2308:
@AayushChaudhary0001 Fixed the issue reported by you. Now, we are hiding the additional fields if the selected project is removed.

webapp/package.json
```diff
@@ -24,7 +24,6 @@
         "react": "16.13.1",
         "react-bootstrap": "1.3.0",
         "react-custom-scrollbars": "4.2.1",
-        "react-dom": "18.2.0",
```
mickmister: Wondering why this was removed. Do you think it's being used by anything? I suppose the bundler would complain if so. I see it's still present in `package-lock.json`, but no package there depends on it

webapp/src/components/modals/create_issue/index.tsx
```diff
@@ -0,0 +1,280 @@
...
+        const {postId, title, channelId} = state[`plugins-${pluginId}` as plugin].createIssueModal;
+        
+        const post = postId ? getPost(state, postId) : null;
+        return {
+            visible: state[`plugins-${pluginId}` as plugin].isCreateIssueModalVisible,
+            post,
+            title,
+            channelId,
+        };
+    });
```
mickmister, line 49: This modal should be split into two components. One should be "should I render right now?" And if yes, it should render the `Modal` component with the other component nested inside. This allows React to clean up state and we don't end up with code like `handleClose` doing a bunch of cleanup.

Same for attach comment modal as well

webapp/src/components/modals/create_issue/index.tsx
```diff
@@ -0,0 +1,280 @@
...
+    useEffect(() => {
+        if (post && !prevPost) {
+            setIssueDescription(post.message);
+        } else if (channelId && (channelId !== prevChannelId || title !== prevTitle)) {
+            setIssueTitle(title.substring(0, MAX_TITLE_LENGTH));
+        }
+    }, [channelId, title, post]);
+
+    // handle issue creation after form is populated
+    const handleCreate = useCallback(async (e: React.FormEvent<HTMLFormElement> | Event) => {
```
mickmister, line 65: Also I recommend not using hooks like `useCallback` for this sort of use case. This modal/component is usually not open/visible when a user is using Mattermost, so it doesn't need to be this optimized. IMO this just adds complexity and risk with all the dependency array stuff, whereas having regular functions does not have any down sides for the purpose of this PR.

The one time it's good to use these is when we are setting up props for a component that will render a lot of items (like a huge array), so it doesn't need to do its heavy lifting rendering all the time. The project selector is potentially like that, but the performance hit on that project select box would first and foremost be the number of DOM elements in the select box if we have a lot of elements. Which we can limit how many are shown ([example from Jira plugin](https://github.com/mattermost/mattermost-plugin-jira/blob/675961b88fa8a23bdb36ec4ee609bf664afb5ec9/webapp/src/components/react_select_setting.tsx#L107-L121)), if we receive complaints when people have over 1000 projects, but I think that's unnecessary for this feature. The only thing that I would say may be worth optimizing are the calls to `useMemo`, but even then, I don't think that's worth it due to the risk it brings.

If there are particular places where you think we need optimizations, please bring them up so we can discuss how to approach it. But I personally think that this feature does not require any of the optimizations.

webapp/src/components/modals/create_issue/index.tsx
```diff
@@ -0,0 +1,280 @@
...
+
+        const dropdownProps = {
+            projectID: project.project_id,
+            projectName: project.name,
+            theme: theme,
+        }
+
+        const labelProps = {
+            selectedLabels: labels,
+            onChange: handleLabelsChange,
```
mickmister, line 145: This is a case where we are not specifying a dependency in the dependency array for this `useMemo` callback. We should supply `handleLabelsChange` here. Even though `handleLabels` change doesn't change, having missing dependencies in the array makes it difficult to refactor in the future, to make sure we don't break anything.

Even if the optimizations were worth it, I think this would be less error-prone as far as dependency arrays:

```tsx
const dropdownProps = useMemo(() => ({
    projectID: project.project_id,
    projectName: project.name,
    theme: theme,
}), [project, theme]);

const labelSelector = useMemo(() => (
    <GitlabLabelSelector
        {...dropdownProps}
        selectedLabels={labels}
        onChange={handleLabelsChange}
    />
), [dropdownProps, labels, handleLabelsChange]);

const issueAttributeSelectors = (
    <>
        {labelSelector}
        // ...
    </>
);
```

But I think that having all of these optimizations makes the code harder to read and is error-prone. And (I know I keep repeating myself), I think for this particular feature the optimizations are not worth the risk that it brings.

webapp/src/components/gitlab_project_selector/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+    const loadProjects = useCallback(async () => {
+        setIsLoading(true);
+        await dispatch(getProjects());
+        setIsLoading(false);
+    }, []);
+
+    const handleOnChange = useCallback((_: string, name: string) => {
+        const project = yourProjects.find((p: Project) => p.path_with_namespace === name);
+        onChange({name, project_id: project?.id});
+    }, [onChange, yourProjects])
```
mickmister, line 44: If the project selector gets "cleared", we should be passing `null` to the parent instead of this object

mickmister (changes requested):
I gave the PR another review. I think the `useMemo`, `useCallback` etc. optimizations may bring more risk than benefit. I recommend we remove them to simplify the code. Please let me know your thoughts

webapp/package.json
raghavaggarwal2308: @ micmister I think it was removed because of duplicity. It's still present on line number `17`.

raghavaggarwal2308:
@mickmister Fixed the review comments added by you.

raghavaggarwal2308:
@mickmister Gentle reminder for re-review

mickmister (commented):
This is looking really good :+1:  I have some more questions/comments for discussion

webapp/src/components/modals/attach_comment_to_issue/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+interface PropTypes {  
+    theme: Theme,
+}
+
+const AttachCommentToIssueModal = ({theme}: PropTypes) => {
+    const [formSubmission, setFormSubmission] = useState<FormSubmission>({
+        isSubmitted: false,
+        isSubmitting: false,
+        error: '',
+    });
```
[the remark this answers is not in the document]
raghavaggarwal2308: @mickmister We are using this state in the "handleClose" function as well. And that function is being used in the modal header so it cannot be moved into the from component.

webapp/src/components/create_issue_form/index.tsx
```diff
@@ -0,0 +1,218 @@
...
+            })
+            return;
+        }
+
+        const postId = post?.id ?? '';
+
+        const issue = {
+            title: issueTitle,
+            description: issueDescription,
+            project_id: project?.project_id,
```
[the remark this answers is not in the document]
raghavaggarwal2308: @mickmister We are validating the project field on line 64.

webapp/src/components/create_issue_form/index.tsx
```diff
@@ -0,0 +1,218 @@
...
+    const prevChannelId = usePrevious(channelId);
+    const prevTitle = usePrevious(title);
+
+    useEffect(() => {
+        if (post && !prevPost) {
+            setIssueDescription(post.message);
+        } else if (channelId && (channelId !== prevChannelId || title !== prevTitle)) {
+            setIssueTitle(title.substring(0, MAX_TITLE_LENGTH));
+        }
+    }, [channelId, title, post]);
```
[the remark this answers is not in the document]
raghavaggarwal2308: > Can you describe when these different cases occur? 
@mickmister This behavior was taken from the Github plugin. 
- When we run the slash command (`/gitlab issue create <title>`), the title value is prefilled.
- When we open the modal from post action menu, the description is prefilled. 

> Seems weird that this form needs to potentially keep track of two posts, when it is only opened in the case of a post menu option or a slash command. I don't think there would be a case where the post "changes" right?
Refactored the code a bit 👍🏽

raghavaggarwal2308:
@mickmister Fixed the review comments added by you

mickmister:
@raghavaggarwal2308 Looks like you may need to push some commits you have locally as I don't see any above. Also, there are conflicts that need to be resolved here

raghavaggarwal2308:
> @raghavaggarwal2308 Looks like you may need to push some commits you have locally as I don't see any above. Also, there are conflicts that need to be resolved here

@mickmister Apologies for that. Updated now.

raghavaggarwal2308:
@mickmister Resolved

raghavaggarwal2308:
@mickmister Gentle reminder for re-review

raghavaggarwal2308:
@mickmister Gentle reminder for re-review

webapp/src/components/issue_attribute_selector.tsx
```diff
@@ -0,0 +1,120 @@
...
+const IssueAttributeSelector = ({isMulti, projectName, theme, label, onChange, loadOptions, selection}: PropTypes) => {
+    const [options, setOptions] = useState<SelectionType[]>([]);
+    const [isLoading, setIsLoading] = useState<boolean>(false);
+    const [error, setError] = useState<string>('');
+
+    useEffect(() => {
+        if (projectName) {
+            loadSelectOptions();
+        }
+    }, []);
```
mickmister, line 32: Should this instead take `projectName` as dep?

```suggestion
    useEffect(() => {
        if (projectName) {
            loadSelectOptions();
        }
    }, [projectName]);
```

webapp/src/components/issue_attribute_selector.tsx
```diff
@@ -0,0 +1,120 @@
...
+    projectName: string;
+    theme: Theme;
+    label: string;
+    onChange: (value: OnChangeType) => void;
+    loadOptions: () => Promise<Array<SelectionType>>,
+    selection: OnChangeType;
+};
+
+const IssueAttributeSelector = ({isMulti, projectName, theme, label, onChange, loadOptions, selection}: PropTypes) => {
+    const [options, setOptions] = useState<SelectionType[]>([]);
```
mickmister, line 24: I can't tell what `SelectionType` is or where it's defined

webapp/src/components/issue_attribute_selector.tsx
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }, []);
+
+    const prevProjectName = usePrevious(projectName);
+
+    useEffect(() => {
+        if (projectName && prevProjectName !== projectName) {
+            loadSelectOptions();
+        }
+    }, [projectName]);
```
mickmister, line 40: Should we add `prevProjectName` to the array?

```suggestion
    useEffect(() => {
        if (projectName && prevProjectName !== projectName) {
            loadSelectOptions();
        }
    }, [projectName, prevProjectName]);
```

mickmister (commented):
Nice work @raghavaggarwal2308 :+1: This feature has a high impact for the plugin

I gave the PR another review and left some comments. Generally approving but want to give you a chance to discuss before approving. Let me know what you think

Also, can you please create some gifs of this feature working so it's easy for UX designers to see it in action?

---

For the project selector in the attach comment modal https://github.com/mattermost/mattermost-plugin-gitlab/pull/306#issuecomment-1493898567

> Just a suggestion for point 4, we can also use the config setting "Gitlab Group". When the setting is empty we can show all issues and when it is specified we can display the issues related to that group only.

Are we able to filter on projects where the user is a member or collaborator, if the gitlab group is not set?

webapp/src/components/issue_attribute_selector.tsx
mickmister: We should really be using the react hooks eslint rule to find these sorts of cases

mickmister:
@raghavaggarwal2308 Are you or another developer planning on continuing work on this?

raghavaggarwal2308:
> @raghavaggarwal2308 Are you or another developer planning on continuing work on this?

@mickmister Yeah, we will work on this ASAP. Apologies for the delay.

webapp/src/components/gitlab_project_selector/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+    useEffect(() => {
+        loadProjects();
+    }, []);
+
+    const loadProjects = async () => {
+        setIsLoading(true);
+        await dispatch(getProjects());
+        setIsLoading(false);
+    };
```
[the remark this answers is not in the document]
ayusht2810: Added error handling
image

webapp/src/components/gitlab_project_selector/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+    addValidate: (key: string, validateField: () => boolean) => void;
+    removeValidate: (key: string) => void;
+};
+
+const GitlabProjectSelector = ({theme, required, onChange, value, addValidate, removeValidate}: PropTypes) => {
+    const [isLoading, setIsLoading] = useState(false);
+
+    const {yourProjects} = useSelector((state: GlobalState) => ({
+        yourProjects: state[`plugins-${manifest.id}` as pluginReduxStoreKey].yourProjects,
+    }));
```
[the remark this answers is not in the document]
ayusht2810: Replaced the `yourProjects` with `myProjects` in the current file. We can update the reducers as part of a separate PR.

webapp/src/components/modals/attach_comment_to_issue/attach_comment_to_issue_form.tsx
```diff
@@ -0,0 +1,132 @@
...
+            comment: message,
+            post_id: post.id,
+            web_url: issueValue?.web_url,
+        };
+
+        setIsSubmitting(true);
+
+        const created = await attachCommentToIssue(comment)(dispatch);
+        if (created.error) {
+            const errMessage = getErrorMessage((created as { error: ErrorType }).error.message);
```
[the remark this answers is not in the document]
ayusht2810: `attachCommentToIssue` function is present in a `js` file https://github.com/Brightscout/mattermost-plugin-gitlab/blob/1f9fbc3912fb9b8f5f7f679dff259cb44a63efa1/webapp/src/actions/index.js#L263 so we cannot currently add type to it.

webapp/src/components/issue_attribute_selector.tsx
ayusht2810: It is defined in the types folder under common types https://github.com/Brightscout/mattermost-plugin-gitlab/blob/1f9fbc3912fb9b8f5f7f679dff259cb44a63efa1/webapp/src/types/common.d.ts#L3

webapp/src/components/form_button.tsx
```diff
@@ -0,0 +1,40 @@
...
+import React from 'react';
+
+type PropTypes = {
+    disabled?: boolean;
+    defaultMessage?: string,
+    btnClass?: string;
+    saving?: boolean;
+    savingMessage?: string;
+    onClick?: () => void;
+};
```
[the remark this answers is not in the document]
ayusht2810: This is used at multiple places, and each uses different props. So none of them are required.

webapp/src/components/sidebar_buttons/sidebar_buttons.tsx
```diff
@@ -19,10 +19,10 @@ interface SidebarButtonsProps {
...
     clientId: string;
     gitlabURL: string;
-    reviews: unknown[];
-    todos: unknown[];
-    yourAssignedPrs: unknown[];
-    yourAssignedIssues: unknown[];
+    reviews?: unknown[];
+    todos?: unknown[];
+    yourAssignedPrs?: unknown[];
+    yourAssignedIssues?: unknown[];
```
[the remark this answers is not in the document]
ayusht2810: This was giving warning in the browser console which got fixed by making them optional.

webapp/src/hooks/use_options.ts
```diff
@@ -0,0 +1,27 @@
+import {useCallback} from 'react';
+import {useDispatch} from 'react-redux';
+
+export const useOptions = (projectName: string, getOptions: FetchIssueAttributeOptionsForProject, returnFields: string[], errorMessage: string, projectID?: number) => {
+    const dispatch = useDispatch();
+
+    const loadOptions = async () => {
+        if (!projectName) {
+            return [];
+        }
```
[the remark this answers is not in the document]
ayusht2810: The assignees and other options are hidden for a user until the user selects the project. Same can be seen in the video.

ayusht2810:
@mickmister Fixed the review fixes. Please re-review.

> Are we able to filter on projects where the user is a member or collaborator, if the gitlab group is not set?

Also can you provide more details on this comment?

mickmister:
> > Are we able to filter on projects where the user is a member or collaborator, if the gitlab group is not set?
> 
> Also can you provide more details on this comment?

@ayusht2810 If the gitlab group is not set in the system console: Can we make it so the user only sees projects where the user a member or collaborator? Meaning, the user shouldn't see projects that they have no association with in the project selectors

webapp/src/types/mattermost-webapp/index.d.ts
```diff
@@ -23,6 +23,8 @@ export interface PluginRegistry {
     registerPopoverUserAttributesComponent(component: React.ReactNode)
     registerLinkTooltipComponent(component: React.ReactNode)
     registerReconnectHandler(handler: any)
+    registerPostDropdownMenuComponent(component: React.ReactNode)
+    registerRootComponent(component: ReactDOM)
```
mickmister, line 27:
```suggestion
    registerRootComponent(component: React.ElementType)
```

webapp/src/types/gitlab_types.d.ts
```diff
@@ -2,9 +2,9 @@ interface IssueBody {
     title: string;
     description: string;
     project_id?: number;
-    labels?: (string | number)[];
-    assignees?: (string | number)[];
-    milestone?: (string | number);
+    labels?: (string | number | Issue)[];
+    assignees?: (string | number | Issue)[];
+    milestone?: (string | number | Issue);
```
mickmister, line 7: Why can all 3 of these be these 3 things? I would think `string` should cover the bases?

webapp/src/components/gitlab_project_selector/index.tsx
```diff
@@ -34,16 +34,25 @@ const GitlabProjectSelector = ({theme, required, onChange, value, addValidate, r
 
     const loadProjects = async () => {
         setIsLoading(true);
-        await dispatch(getProjects());
+        const res: any = await dispatch(getProjects());
```
mickmister, line 37: We want to avoid `any` when possible

```suggestion
        const res = (await dispatch(getProjects())) as {error?: ErrorType};
```

webapp/src/components/form_button.tsx
```diff
@@ -3,16 +3,16 @@
 
 import React from 'react';
 
-type PropTypes = {
+type Prop = {
```
mickmister, line 6:
```suggestion
type Props = {
```

mickmister (approved):

raghavaggarwal2308:
@mickmister Fixed the review comments added by you

webapp/src/components/modals/create_issue/create_issue_form.tsx
```diff
@@ -62,9 +62,9 @@ const CreateIssueForm = ({theme, handleClose, isSubmitting, setIsSubmitting}: Pr
             title: issueTitle,
             description: issueDescription,
             project_id: project.project_id,
-            labels: labels.map((label) => label.value),
-            assignees: assignees.map((assignee) => assignee.value),
-            milestone: milestone?.value,
+            labels: labels.map((label) => label.value as string),
+            assignees: assignees.map((assignee) => assignee.value as number),
+            milestone: milestone?.value as number,
```
mickmister, line 67: Can we instead have these typed separately, rather than the `string | number | Issue` thing? Then we wouldn't need `as` here I think. It's fine this is the easiest way, just curious if it needs to be this way

mickmister (approved):

ayusht2810:
@mickmister Fixed the review fixes. Please re-review

mickmister (approved):

ayusht2810:
> Expected: Private projects should not be accessibile without the permission

@AayushChaudhary0001 Regarding the expected behavior, the system console settings for `Enable Private Repositories` is related to subscription and not to this issue creation feature as seen in the help text `(Optional) Allow the plugin to work with private repositories for subscriptions.`  Should we update the help text to disable private repo for all the features if the setting is disabled?

AayushChaudhary0001 (approved):
Approving this PR, as everything is working fine for this, will create a separate issue regarding the flow above.

PR #302, opened 2022-06-07
Add support for triggering pipelines and job status updates

spirosoik:
#### Summary
Add the capability to trigger the Gitlab pipelines. In parallel after the trigger if you are subscribed to jobs you will get status updates for jobs of the pipelines.dc

#### Ticket Link
Ticket: #213 & #289

PR #298, opened 2022-05-26
Fixed the Oauth token expiring issue (issue #275)

manojmalik20:
This PR fixes issue #275.

Starting with GitLab 15.0.0,  OAuth tokens always expire after 2 hours. Currently, this plugin doesn't refresh the token after it is expired. So, I added the logic of refreshing the OAuth token if expired and using the new token to create the Gitlab client.

#### Related links
https://github.com/mattermost/chimera/pull/9

PR #308, opened 2022-07-10
[GH-261] Handle revoked token

MatthewDorner:
#### Summary
If a GitLab API call fails with `401 {error: invalid_token}`, disconnect the user and DM the user a message telling how to reconnect their account.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-gitlab/issues/261

Questions:
- I have not handled the case where the token is both expired and revoked. Here, `p.checkAndRefreshToken` attempts to refresh the token and the call to `src.Token` fails. However, if I handle the error and call `p.disconnectGitlabAccount`, it will call `p.checkAndRefreshToken` again, creating an infnite loop. How would you suggest to resolve this?

- In the case where the error occurs during the execution of a slash command, I have not changed the existing slash command error responses. I just send the DM in addition. So if you get the error during `/gitlab me`, you'll get the new message in the DM channel but the command response will still be "Encountered an error getting your GitLab profile.". Should I change the command response to a similar or identical message to what I'm sending in the DM?

- What test coverage is needed?


server/api.go
```diff
@@ -516,6 +516,9 @@ func (p *Plugin) getConnected(c *Context, w http.ResponseWriter, r *http.Request
 func (p *Plugin) getUnreads(c *UserContext, w http.ResponseWriter, r *http.Request) {
 	result, err := p.GitlabClient.GetUnreads(c.Ctx, c.GitlabInfo)
 	if err != nil {
+		if strings.Contains(err.Error(), invalidTokenError) {
+			p.handleRevokedToken(c.GitlabInfo)
+		}
```
mickmister, line 521: Instead of duplicating this block across the different functions, I'm thinking we should have a function `handleGitlabError(err error)`, which has the above logic. What do you think @MatthewDorner?

server/command.go
```diff
@@ -511,11 +520,15 @@ func (p *Plugin) subscriptionsAddCommand(ctx context.Context, info *gitlab.UserI
...
 	if err != nil {
-		if errors.Is(err, gitlab.ErrNotFound) {
+		switch {
+		case errors.Is(err, gitlab.ErrNotFound):
 			return "Resource with such path is not found."
-		} else if errors.Is(err, gitlab.ErrPrivateResource) {
+		case errors.Is(err, gitlab.ErrPrivateResource):
 			return "Requested resource is private."
+		case strings.Contains(err.Error(), invalidTokenError):
+			p.handleRevokedToken(info)
```
mickmister, line 529: The code cleanup is good here, though I think for the purpose of this PR, can we just add a call to `handleGitlabError(err)` at the top of the `if err != nil` block?

server/plugin.go
```diff
@@ -618,3 +628,12 @@ func (p *Plugin) checkAndRefreshToken(token *oauth2.Token) (*oauth2.Token, error
 
 	return nil, nil
 }
+
+func (p *Plugin) handleRevokedToken(info *gitlab.UserInfo) {
+	p.disconnectGitlabAccount(info.UserID)
+	err := p.CreateBotDMPost(info.UserID, "Your GitLab account was disconnected due to an invalid or revoked authorization token. Reconnect your account using the `/gitlab connect` command.", "custom_git_revoked_token")
```
mickmister, line 634: Since this isn't a custom post that the webapp plugin is going to render, we can leave the `postType` out of this function call:

```suggestion
	err := p.CreateBotDMPost(info.UserID, "Your GitLab account was disconnected due to an invalid or revoked authorization token. Reconnect your account using the `/gitlab connect` command.", "")
```

mickmister (changes requested):
Awesome work on this @MatthewDorner, thanks for implementing this solution :+1:

The PR mostly looks good. I have one request to make a common function `handleGitlabError` to avoid some code duplication

mickmister:
> * I have not handled the case where the token is both expired and revoked. Here, `p.checkAndRefreshToken` attempts to refresh the token and the call to `src.Token` fails. However, if I handle the error and call `p.disconnectGitlabAccount`, it will call `p.checkAndRefreshToken` again, creating an infnite loop. How would you suggest to resolve this?

What is the side effect of "not handling the case where the token is both expired and revoked"? Note that the plugin's API is called multiple times when the frontend requests data to fill the left hand side GitLab metrics, so it will always be tried more than once during this time. You can refresh your browser to trigger this data refresh.

There may be the case where we end up sending more than one bot DM here, due to more than one request trying to use the access token before it is removed from the kv store. If the server is running in a high availability environment, there may be some lag on the database read for rejecting the token, which could cause this scenario.

I'm thinking that sending multiple DMs is an ok outcome, as it's relatively harmless, and any defense around this is going to be complicated and potentially error-prone.

Also, we will want to make sure the changes in this PR work with the changes from https://github.com/mattermost/mattermost-plugin-gitlab/pull/298

> * In the case where the error occurs during the execution of a slash command, I have not changed the existing slash command error responses. I just send the DM in addition. So if you get the error during `/gitlab me`, you'll get the new message in the DM channel but the command response will still be "Encountered an error getting your GitLab profile.". Should I change the command response to a similar or identical message to what I'm sending in the DM?

What you describe as the current behavior of this PR seems good to me :+1:

> * What test coverage is needed?

I would test the `handleGitlabError` function to assert that it handles the error properly

server/plugin.go
MatthewDorner: OK. I will note that there are other DM posts which specify custom post types such as `custom_git_welcome`, `custom_git_todo`. But I don't see where the webapp contains definitions for any of these. I was trying to just make it follow that convention.

server/command.go
MatthewDorner: Yeah I'll do that.

server/api.go
MatthewDorner: Yes, I think that would be better.

MatthewDorner:
> What is the side effect of "not handling the case where the token is both expired and revoked"? Note that the plugin's API is called multiple times when the frontend requests data to fill the left hand side GitLab metrics, so it will always be tried more than once during this time. You can refresh your browser to trigger this data refresh.

Each API call will get to `checkAndRefreshToken`, see that the token needs to be refreshed, but when it tries to refresh, the call to `src.Token()` will fail because the token is revoked and can't be refreshed. So each request will fail. This is the current behavior in this PR if the token is both expired and revoked. And I'm not sure what exact circumstances a user would revoke a token, but it seems like it might coincide with an expired token (they stop using the plugin for a while or something) so that this can be a realistic scenario.

MatthewDorner:
> Also, we will want to make sure the changes in this PR work with the changes from https://github.com/mattermost/mattermost-plugin-gitlab/pull/298

Yeah, I waited for that PR to be merged and this PR is based on code that includes those changes. But the issue with the "revoked and expired" case is how to make this work with the `checkAndRefreshToken` method from that PR. To clarify:
- currently in this PR, the case is not handled which means all requests will fail if the token is revoked and expired
- if I try to handle the error from the `src.Token()` call within `checkAndRefreshToken` the same way I have elsewhere, it will cause an infinite loop in the code due to the reasons described in my opening post

There is probably a workaround I'm just not seeing?

server/api.go
MatthewDorner: I guess the function would need to take a `*gitlab.UserInfo` as well. Maybe the `handleGitlabError` should be combined with my `handleRevokedToken` as a single function such as `checkAndHandleRevokedToken`?

mickmister:
> > What is the side effect of "not handling the case where the token is both expired and revoked"? Note that the plugin's API is called multiple times when the frontend requests data to fill the left hand side GitLab metrics, so it will always be tried more than once during this time. You can refresh your browser to trigger this data refresh.
> 
> Each API call will get to `checkAndRefreshToken`, see that the token needs to be refreshed, but when it tries to refresh, the call to `src.Token()` will fail because the token is revoked and can't be refreshed. So each request will fail. This is the current behavior in this PR if the token is both expired and revoked. And I'm not sure what exact circumstances a user would revoke a token, but it seems like it might coincide with an expired token (they stop using the plugin for a while or something) so that this can be a realistic scenario.

@MatthewDorner Is there any risk with this situation? If it's revoked & expired, and this happens, is the kvstore then in an invalid state? It seems like a non-issue to me if we invalidate the token on our side so we don't try to use it again.

> > Also, we will want to make sure the changes in this PR work with the changes from #298
> 
> Yeah, I waited for that PR to be merged and this PR is based on code that includes those changes. But the issue with the "revoked and expired" case is how to make this work with the `checkAndRefreshToken` method from that PR. To clarify:
> 
> * currently in this PR, the case is not handled which means all requests will fail if the token is revoked and expired
> * if I try to handle the error from the `src.Token()` call within `checkAndRefreshToken` the same way I have elsewhere, it will cause an infinite loop in the code due to the reasons described in my opening post
> 
> There is probably a workaround I'm just not seeing?

"This case is not handled"

There's something I'm missing here. I don't understand *how* the case is not handled. I understand that the token won't work in this case. Are we getting the "invalid token" error from GitLab in this case? Shouldn't we just delete it from our database and tell the user about it like usual? Is there something stopping us from doing this? Can we extract some code from the function causing the infinite loop to call in another context?

Also, please mention me in replies here so I get a notification for the mention :+1:

MatthewDorner:
In `checkAndRefreshToken`, we just check the expiration time of the token from the KV store to determine if it needs to be refreshed. If a token is invalid, we only know when we make a request and it returns the error from GitLab. So in order to reproduce this condition:
- shut down Mattermost server
- wait 2 hours, or modify the `expiry` field in the KV store `_gitlabtoken` record
- on the GitLab side, revoke the token
- start server again

From my perspective, I feel like the issue is that `checkAndRefreshToken` from https://github.com/mattermost/mattermost-plugin-gitlab/pull/298 is called from `getGitlabUserInfoByMattermostID`, which ensures it's called before every API call, but isn't really part of the purpose of that function. And it means `checkAndRefreshToken` can't then use those methods such as in my case `p.disconnectGitlabAccount` without causing a loop.

I can think about it more, just wanted to see if something obvious occurred.

MatthewDorner:
>  Are we getting the "invalid token" error from GitLab in this case? Shouldn't we just delete it from our database and tell the user about it like usual?

@mickmister We never get to the actual API call, as the call to `src.Token()` (calls GitLab to refresh the token since it's expired) in `checkAndRefreshToken` will fail and that error will be returned. If I add code in `checkAndRefreshToken` to check for that error and eventually call `p.disconnectGitlabAccount`, that will call `getGitlabUserInfoByMattermostID` which will call `checkAndRefreshToken`, which will again try to refresh the token and on and on. Sorry if it's confusing.

mickmister:
> > Are we getting the "invalid token" error from GitLab in this case? Shouldn't we just delete it from our database and tell the user about it like usual?
> 
> @mickmister We never get to the actual API call, as the call to `src.Token()` (calls GitLab to refresh the token since it's expired) in `checkAndRefreshToken` will fail and that error will be returned. If I add code in `checkAndRefreshToken` to check for that error and eventually call `p.disconnectGitlabAccount`, that will call `getGitlabUserInfoByMattermostID` which will call `checkAndRefreshToken`, which will again try to refresh the token and on and on. Sorry if it's confusing.

@MatthewDorner Yeah it would seem to me that `getGitlabUserInfoByMattermostID` should be a local operation, and not a "contact GitLab" operation. `getGitlabUserInfoByMattermostID` should not call `checkAndRefreshToken`, and any function calling `getGitlabUserInfoByMattermostID` that also cares about refreshing the token should explicitly call `checkAndRefreshToken` after the call `getGitlabUserInfoByMattermostID` has returned. Does this solve the issue?

MatthewDorner:
@mickmister yes, that is one of the solutions that occurred to me. Basically, check for both conditions (expired and revoked) in the same place.

I didn't want to go that route immediately since it would be a larger scope of changes than I did so far, so I will work on it a bit and update. Thanks for the input.

MatthewDorner:
@mickmister Think it would work to use a wrapper or decorator around the GitLab client calls? Otherwise, every method that makes GitLab calls must call `checkAndRefreshToken` in its setup stage, and then later check for the invalid token after any call it makes, or at least after the first call it will definitely make, even if it's only to call `handleGitlabError(err error)`.

A wrapper function should be cleaner and use less code, but I assume it would hit the KV store for every GitLab call rather than only once per method that makes GitLab calls, but that may be negligible. And may be difficult in other ways I don't foresee.

mickmister:
@MatthewDorner Could we have a wrapper function around the http handler functions?

https://github.com/mattermost/mattermost-plugin-gitlab/blob/c5d107ec3962fd8ba53596d595c703922bd6f3e3/server/api.go#L52

We can rename `checkAuth` to `checkMattermostAuth`, then make a similar `checkGitlabAuth` function for certain routes.

mickmister:
> @mickmister Think it would work to use a wrapper or decorator around the GitLab client calls? Otherwise, every method that makes GitLab calls must call `checkAndRefreshToken` in its setup stage, and then later check for the invalid token after any call it makes, or at least after the first call it will definitely make, even if it's only to call `handleGitlabError(err error)`.
> 
> A wrapper function should be cleaner and use less code, but I assume it would hit the KV store for every GitLab call rather than only once per method that makes GitLab calls, but that may be negligible. And may be difficult in other ways I don't foresee.

I'm not sure how this wrapper function would work, since each of the Gitlab client's methods return different types. To me, `handleGitlabError` is the wrapper function. We would probably need to abstract out the Gitlab client to include this wrapper behavior.

If the solution in my previous comment won't work, I'm thinking we should just go with the strategy designed in your first paragraph. Explicit error handling for each call to the external library Gitlab client makes sense to me.

MatthewDorner:
@mickmister unless you have a strong preference, I think I like locating the GitLab token error handling closer to the GitLab client calls, and to ensure it will be called in more circumstances such as slash commands, even if that isn't strictly necessary due to the frontend constantly making those requests. Also I already started working on that version haha.

MatthewDorner:
> I'm thinking that sending multiple DMs is an ok outcome, as it's relatively harmless, and any defense around this is going to be complicated and potentially error-prone.

@mickmister yes, this will be the case when the revoked token is detected due to the 4 API requests from the frontend. It also spams the error log a bit with messages such as `can't get GitLab user info from mattermost id` because the user is already disconnected and it's still trying to process the other requests.

MatthewDorner:
@mickmister Converting to draft because still having issues. The issue is related to the conversation https://github.com/mattermost/mattermost-plugin-gitlab/pull/298#discussion_r886170548 and case where the 4 API requests happen simultaneously, as well as the previous conversation in this PR about token being both expired and revoked.

While it was OK to have multiple token refreshes in the PR linked above, it is not OK once we are checking for and disconnecting the user based on an invalid token. What happens is one of the 4 requests will refresh the token, and the old token is now invalid. Then the 3 other goroutines try to refresh their (now invalid) tokens and (with the changes in this PR) the user gets disconnected. I'm not sure how the exact timing or concurrency works out, but this condition happens every time for me when I set my token to expire and then refresh the browser.

This is why it's a problem if `getGitlabUserInfoByMattermostId` is a local operation: one goroutine can load the token from KV while another has already refreshed it and is about to store it, which means the first goroutine has loaded an invalid token.

I suspect this is also the reason for a user in the community server complaining of their error logs being spammed even after the token refresh fix, which resulted in this issue https://github.com/mattermost/mattermost-plugin-gitlab/issues/313. The first request would refresh the token and the other 3 would fail and write to the error log, every 2 hours. It just wouldn't break any functionality, but if we also disconnect the user every time we detect an invalid token it becomes a bigger issue, and maybe some concurrency control is required now.

MatthewDorner:
The issue is made more difficult by the fact that GitLab's OAuth implementation does not allow a `refresh_token` to be used multiple times. It issues a new one upon each refresh. I guess the OAuth specification allows it to be done either way.

https://docs.gitlab.com/ee/api/oauth2.html

The Golang `oauth2` library apparently auto-refreshes the token, but doesn't give you the new token back, so the auto-refresh doesn't work in cases like this where `refresh_token` is invalidated in this way and the token needs to be persistent. (and would be subject to the same concurrency issues I suppose). Anyway by detecting the need to refresh the token ahead of time we're avoiding that auto-refresh behavior.

Relevant thread here, although it's confusing because they're talking about a few different issues / cases at the same time:
https://github.com/golang/oauth2/issues/84

mickmister:
> The Golang `oauth2` library apparently auto-refreshes the token, but doesn't give you the new token back

Doesn't the current solution for auto-refreshing the token in this project rely on us receiving a new version of the token from the library?

MatthewDorner:
@mickmister We are refreshing the token manually. That auto refresh is a separate feature that's discussed in the thread I linked but I don't think it works here since the GitLab Oauth provider invalidates the refresh token, thus why we had to make the other recent token fix.

mickmister:
@MatthewDorner Sorry it's difficult for me to grasp based on our discussion here. What are the remaining items to resolve here? This PR adds a lot of value to the plugin, so I would like to see this merged 🙂

If the `oauth2` library doesn't "give you back" the token when it's refreshed, does it at least mutate the existing one so it's updated?

mickmister:
To solve the issue of 4 requests being done simultaneously, maybe the UI can issue a single request first for this purpose, and then follow up with the 4 requests to fetch data afterwards. But we also have the issue of having multiple tabs open simultaneously, including the desktop app which always has 3 tabs open due to the different products supported now (Playbooks and Boards).

mickmister:
If it's an absolute requirement that we refresh only once per user, we can also use the `cluster.Mutex` function exposed [here](https://github.com/mattermost/mattermost-plugin-api/blob/73008cfda2fbefe305f704eeb45975fabe39d7f2/cluster/mutex.go#L41). For the mutex's key, we can use `userID + "-refresh-token"`. This specific mutex is necessary to coordinate between different instances of the plugin running on separate nodes of a high availability cluster.

If we make it so the frontend always sends a request to a new plugin endpoint `/token-refresh` before fetching data, we can use the mutex to ensure only one instance of the plugin is checking the token at a given time. This allows us to make sure the token is only refreshed one time. @MatthewDorner What do you think of this?

cc @DHaussermann for visibility

MatthewDorner:
@mickmister @DHaussermann 

Using `cluster.Mutex` fixed the concurrency issue. As a heads up, I found a lot of instances of other people having this problem with OAuth providers that implement a strategy called "refresh token rotation" which seems to have gotten popular recently. Here is another example of the issue: https://github.com/dexidp/dex/pull/2352. So, this may appear in other plugins.

As for the potential `/token-refresh` endpoint, it could clean up the code so we don't have to check for expired token in each individual API handler on the backend, but the mutex by itself is what solves the concurrency issue. There are also downsides to adding the endpoint, as it adds responsibility onto the frontend and the API, so I'd like to consider and get input on what could be the best solution. Functionally, I think the existing solution suffices.

Also I see the test is failing. I think it's just something in the test code that needs to be corrected for my changes, looking into it now.

mickmister:
> There are also downsides to adding the endpoint, as it adds responsibility onto the frontend and the API

I think we're good with the mutex on the backend, no need to add the complexity involving the frontend.

At the moment, it seems the backend is a bit bloated with the code duplication from the PR. Can we move the code to a middleware function, similar to `p.checkAuth`? Then we can apply the new middleware to the appropriate routes here https://github.com/mattermost/mattermost-plugin-gitlab/blob/2c3003baec9efac58d02299cd1ee2d29a7c2c08c/server/api.go#L52

mickmister:
> Also I see the test is failing. I think it's just something in the test code that needs to be corrected for my changes, looking into it now.

It looks like the `api.go` file needs to be formatted with `gofmt`. My editor does this automatically whenever I save a file. That's strange if that's not the case for your editor

```
golangci-lint run ./...
server/api.go:112: File is not `gofmt`-ed with `-s` (gofmt)
		mutex, err := cluster.NewMutex(p.API, context.UserID + "-refresh-token")
make: *** [Makefile:45: check-style] Error 1
```

MatthewDorner:
> At the moment, it seems the backend is a bit bloated with the code duplication from the PR. Can we move the code to a middleware function, similar to `p.checkAuth`? Then we can apply the new middleware to the appropriate routes here

@mickmister OK, added `checkToken`

MatthewDorner:
@mickmister
- Should we also have the mutex for the `/connected` endpoint? It doesn't use `attachUserContext` which is where I put the mutex code. But, it sometimes calls `GetToDo` which calls GitLab, so it could need the mutex but I'd be adding it to the api handler itself. This seems to be an issue where `/connected` is serving two separate functions.
- Similarly for the '/todo' endpoint, I didn't use the `checkToken` wrapper there since I already put a call to `p.checkAndRefreshToken` in `GetToDo` since at that time I was trying to put the token checks as close as possible to the GitLab client calls. I could instead refactor so all the calls to `p.checkAndRefreshToken` (there are 3) are moved out of the `server/plugin.go` methods and instead are placed in `api.go` and `command.go`. Is there a preference?

mickmister:
@spirosoik Heads up on syncing changes from https://github.com/mattermost/mattermost-plugin-gitlab/pull/302 here. Any user action that has to do with interacting with GitLab's API needs to handle token expiration before the call occurs. This is due to changes in how GitLab handles token expiration/refresh. This is also the reason we have the plugin disabled on community currently.

server/plugin.go
```diff
@@ -599,22 +598,51 @@ func (p *Plugin) HasGroupHook(ctx context.Context, user *gitlab.UserInfo, namesp
...
 
-func (p *Plugin) checkAndRefreshToken(token *oauth2.Token) (*oauth2.Token, error) {
+func (p *Plugin) checkAndRefreshToken(userInfo *gitlab.UserInfo) error {
 	// If there is only one minute left for the token to expire, we are refreshing the token.
 	// The detailed reason for this can be found here: https://github.com/golang/oauth2/issues/84#issuecomment-831492464
 	// We don't want the token to expire between the time when we decide that the old token is valid
 	// and the time at which we create the request. We are handling that by not letting the token expire.
-	if time.Until(token.Expiry) <= 1*time.Minute {
+
+	if time.Until(userInfo.Token.Expiry) <= 1*time.Minute {
```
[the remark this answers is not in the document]
mickmister: Also, we need to load the token from the kv store while we have the mutex locked, otherwise we aren't synchronizing access to the token.

mickmister (changes requested):
Thanks for your continued work on this @MatthewDorner. I have a few comments on clean up, and moving the mutex functionality to the `checkAndRefresh` function.

server/api.go
```diff
@@ -52,10 +52,10 @@ func (p *Plugin) initializeAPI() {
 
 	apiRouter.HandleFunc("/user", p.checkAuth(p.attachContext(p.getGitlabUser), ResponseTypeJSON)).Methods(http.MethodPost)
 	apiRouter.HandleFunc("/todo", p.checkAuth(p.attachUserContext(p.postToDo), ResponseTypeJSON)).Methods(http.MethodPost)
```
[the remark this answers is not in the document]
mickmister: > * For the '/todo' endpoint, I didn't use the `checkToken` wrapper there since I already put a call to `p.checkAndRefreshToken` in `GetToDo` since at that time I was trying to put the token checks as close as possible to the GitLab client calls. I could instead refactor so all the calls to `p.checkAndRefreshToken` (there are 3 in `server/plugin.go`) are moved out of the `server/plugin.go` methods and instead are placed in `api.go` and `command.go`. Is there a preference?

Sorry I read this but still made my comment above. Since we may be going with the strategy of checking at the top of each function that uses the gitlab client, I think having this in `GetToDo` works well.

mickmister:
> * Should we also have the mutex for the `/connected` endpoint? It doesn't use `attachUserContext` which is where I put the mutex code. But, it sometimes calls `GetToDo` which calls GitLab, so it could need the mutex but I'd be adding it to the api handler itself. This seems to be an issue where `/connected` is serving two separate functions.

If we are going with the strategy of moving the functionality out of the middleware, I think having the check in `GetToDo` handles this, so no issues here.

server/plugin.go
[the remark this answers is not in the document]
MatthewDorner: The mutex is in `attachUserContext` because I wanted to lock it right before calling `getGitlabUserInfoByMattermostID`.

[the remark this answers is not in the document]
MatthewDorner: I'm not sure how it will work to move the mutex to the `checkAndRefreshToken` function, as I need to lock it before calling `GetGitlabUserInfoByMattermostId`, and only unlock it after the logic in `checkAndRefreshToken`, right?

If you're suggesting reverting the `checkToken` function I added most recently and going back to having more per-function calls to `checkAndRefreshToken`, that's fine, but I'm not sure about the mutex.

Can you lay out how you see it working for one request, where do you see each of: acquiring the mutex lock, loading the token from KV, check and refresh and save to KV, and unlock mutex?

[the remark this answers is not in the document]
MatthewDorner: Are you suggesting going closer to the original solution implemented in the previous token fix PR (the current `master` code), where `checkAndRefreshToken` is just called from `GetGitlabUserInfoByMattermostId`? We could do that, coding a workaround for the loop issue I was having (described earlier in the thread) would be simple enough given how much work it's ended up to do it another way. Something like a function that loads the user info without checking the token (since that was causing the loop) which could be used by `p.DisconnectGitlabAccount`.

server/api.go
```diff
@@ -108,7 +109,17 @@ func (p *Plugin) attachUserContext(handler HTTPHandlerFuncWithUserContext) http.
...
+		mutex, err := cluster.NewMutex(p.API, context.UserID+"-refresh-token")
+		if err != nil {
+			p.writeAPIError(w, &APIErrorResponse{ID: "", Message: "Unable to get mutex: " + err.Error(), StatusCode: http.StatusInternalServerError})
+			return
+		}
+
+		mutex.Lock()
+
 		info, apiErr := p.getGitlabUserInfoByMattermostID(context.UserID)
+		defer mutex.Unlock()
```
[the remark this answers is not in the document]
MatthewDorner: @mickmister It would be necessary to lock the entire handler if the call to `checkAndRefreshToken` is inside the handler, as it was up until my most recent change which moved it to the `checkToken` wrapper instead. So I could probably work it so `attachUserContext` and `checkToken` are combined somehow and the lock is released before the handler is called, except it sounds in your other comments like you are thinking of putting the token check logic back inside the handlers?

Yeah I'm not sure why this problem is so difficult, but I've been kind of going back and forth between these options myself: should the check be closer to the GitLab calls, or should it be closer to the origin of the request?

MatthewDorner:
@mickmister Updated by moving the mutex `defer` statement as requested. Added the `requireConnected` parameter to `attachUserContext` as discussed in video call with @mickmister. The reasoning here is that `getConnected` needs to be locked by the mutex but we only want to add the mutex logic in one place since it's sensitive, so this change is to allow `getConnected` to use `attachUserContext`. Could be improved but this will allow the plugin to be fixed in the meantime. For similar reasons, the lock is being left covering the entire API handler.

mickmister:
@levb Please take a look at this if you have the chance. This is currently blocking usage of this plugin on GitLab cloud, and the last few versions of on-prem GitLab

server/api.go
```diff
@@ -103,21 +104,32 @@ type UserContext struct {
...
-func (p *Plugin) attachUserContext(handler HTTPHandlerFuncWithUserContext) http.HandlerFunc {
+func (p *Plugin) attachUserContext(handler HTTPHandlerFuncWithUserContext, requireConnected bool) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		context, cancel := p.createContext(w, r)
 		defer cancel()
 
-		info, apiErr := p.getGitlabUserInfoByMattermostID(context.UserID)
-		if apiErr != nil {
-			p.writeAPIError(w, apiErr)
+		mutex, err := cluster.NewMutex(p.API, context.UserID+"-refresh-token")
```
levb, line 112: I am confused, is the idea to obtain a mutex for each request to the API, to avoid racing to refresh the token?

MatthewDorner: Yes, to make it compatible with the GitLab OAuth provider using the "refresh token rotation" strategy.

mickmister:
@MatthewDorner After talking it over with Lev, we arrived at a proposed solution to use a wrapper function to access the GitLab clients on demand. I think this is one of the original solutions you proposed so I apologize how it went through a couple revisions and circled back to this. This will mean that the token will only be examined/used at the point of the client usage. One of the goals here is to claim the mutex only when we absolutely need to.

I'm thinking it would work something like this:

```go
func useGitlabClient(userID string, toRun func(client gitlab.Client)) error {
	userInfo := fetchUserInfo(userID)
	token := userInfo.token
	if token.expired {
		mutex := aquireMutex()
		mutex.Lock()
		defer mutex.Unlock()

		userInfo = fetchUserInfo()
		token = refreshToken(token)
	}

	client := makeClient(token)
	err := toRun(client)
    if err != nil {
        if err.Error() != tokenError {
            return err
        }

        mutex := aquireMutex()
        mutex.Lock()
        defer mutex.Unlock()

        userInfo = fetchUserInfo(userID)
        token = refreshToken(token)
    }

	return toRun(client)
})

// Then an example of using this function:

func getPullRequests(userID string) (PullRequest[], error) {
	var prs PullRequest[]

	err := useGitlabClient(userID, func (client gitlab.Client) error {
		prsLocal, err := client.GetPullRequests()
		if err != nil {
			return err
		}

		prs = prsLocal
	})

	if err != nil {
		return nil, err
	}

	return prs, nil
}
```

What do you think about this? I think it ends up with much cleaner code to understand and to manage functionally

MatthewDorner:
@mickmister If we check the token for expiration before getting the mutex, that means we loaded the token already and that first call to `userInfo := fetchUserInfo(userID)` is not locked. So, I don't think the mutex will do anything to stop another goroutine from running that code and loading the same token while the first goroutine is refreshing it, which means it wouldn't fix the issue?

MatthewDorner:
And you're getting around the fact that client returns different types for each method by using this closure type thing?

I'm trying to see where the client comes from here, your `makeClient(token)` would be like `func (g *gitlab) gitlabConnect(token oauth2.Token) (*internGitlab.Client, error)` which is in the code?

mickmister:
If we load the token without mutex, and it's not expired, I think we can assume it's a valid token to use. We should also recheck the expiration right after grabbing the mutex, before refreshing. This is because another goroutine could have refreshed it since the previous expire check.

The code I wrote is error prone actually, because we would end up grabbing the mutex twice potentially. Those need to be done in an inline function so we can properly defer the unlock.

mickmister:
> I'm trying to see where the client comes from here, your `makeClient(token)` would be like `func (g *gitlab) gitlabConnect(token oauth2.Token) (*internGitlab.Client, error)` which is in the code?

Yeah that looks right, just used as a callback mechanism now. We would likely get rid of attachUserContext, and use just attach context instead

MatthewDorner:
> If we load the token without mutex, and it's not expired, I think we can assume it's a valid token to use.

It's the same assumption we already make. We check if it's going to expire within the next minute, and if so refresh it, so it's not an issue of it expiring while the packet is going out or whatever. I don't know how it will fare in the situation where one request doesn't need to refresh, and instantaneously after that the next request decides it does need to refresh. I assume that by reloading the token from KV store before we try to refresh it, would make it unlikely that the "refresh" request somehow ends up going out before the first one.

MatthewDorner:
> The code I wrote is error prone actually, because we would end up grabbing the mutex twice potentially. Those need to be done in an inline function so we can properly defer the unlock.

I don't think you need the second part with the mutex there. If the request fails, it shouldn't be due to an expired token because we will have always checked beforehand. The current code also makes this assumption which I think is fair.

MatthewDorner:
@mickmister Would `getPullRequests` in your example be what we'd do to `func (g *gitlab) GetYourPrs(ctx context.Context, user *UserInfo) ([]*internGitlab.MergeRequest, error)` in `server/gitlab/api.go`? Those methods take a `*UserInfo` already so might not have to load it the initial time as in the example.

mickmister:
I was just writing pseudo code. In reality, that would be the http handler. I suppose the `gitlab` struct can become an instance that already has the user info attached, then its method parameters would no longer rewuire

MatthewDorner:
If it goes in the http handlers in `server/api.go`, it will only check the token for calls that originate from API requests. There are other places like `func (p *Plugin) GetToDo(ctx context.Context, user *gitlab.UserInfo) (bool, string, error)` which makes the same call (to get your PRs) to GitLab and various other uses.

It would be good if it could be integrated into the methods in `server/gitlab/api.go`, then all the code in `main` package wouldn't have to deal with it.

MatthewDorner:
@mickmister I guess we can't put it in `gitlab` package, because we can't access the KV store inside `gitlab` because we don't have access to the `main.Plugin`. Well, I was trying to write something like:

```
func (p *Plugin) withValidToken(toRun func(g *gitlab.Gitlab, ctx context.Context, user *gitlab.UserInfo, args ...any) (any, error), ctx context.Context, user *gitlab.UserInfo, args ...any) (any, error) {

	if time.Until(user.Token.Expiry) <= 1*time.Minute {
		var mutex sync.Mutex
		mutex.Lock()
		defer mutex.Unlock()

		user, _ = p.getGitlabUserInfoByMattermostID(user.UserID)
		p.checkAndRefreshToken(user)
	}

	return toRun(ctx, user, args)
}
```

that could go in `main` and you could use like:

`gitUser, err := p.withValidToken(p.GitlabClient.GetUserDetails, ctx, info)`

Which I can't get to work either because of the fact that `gitlab` methods have different function signatures and return types, or maybe because I'm trying to pass it a method instead of just a regular function? But I thought there must be some way around those issues. Anyway I guess it doesn't save that much code compared to your example. It's just that with yours, you need to write the callback every time so you'll end up duplicating it if `p.GitlabClient.GetUserDetails` for example is used in more than one place.

mickmister:
> You also need to write separate callbacks for each function, it would be nice if it could be a simple wrapper like this that returns a generic type or whatever, I don't know enough about Golang typing but thought that should be possible somehow.

The "simple wrapper like this that returns a generic type" isn't possible with Go, without jumping through a bunch of hoops of type assertions which will just make the code more complicated in this case.

If the callback solution doesn't seem appropriate, I'm thinking the `gitlab` client can be modified to be user-specific, by storing the user's token as a struct field. Then it can be instantiated as needed instead of being a singleton on the `Plugin` struct. When the client is being instantiated, we can perform any necessary token refreshes. Does this make sense?

MatthewDorner:
I started implementing based on your pseudocode. Some things I found:
- it requires to import `github.com/xanzy/go-gitlab` from the `main` package so you can define the result variable that you are assigning to from the callback / closure such as `result := []*gitlab.MergeRequest{}`, maybe not a big deal
- I don't think it will work for the `getConnected` handler, or any case where we might load a user data, then make a client call (which itself might refresh and re-save the user data) and then save the user data (which would now include the old invalid token). This was an issue for my solution as well where we ended up using a workaround. Mainly the issue is storing the token as part of user data in the KV store. https://github.com/mattermost/mattermost-plugin-gitlab/blob/2c3003baec9efac58d02299cd1ee2d29a7c2c08c/server/api.go#L476

> I'm thinking the gitlab client can be modified to be user-specific, by storing the user's token as a struct field. Then it can be instantiated as needed instead of being a singleton on the Plugin struct. 

Will this have the same issue as above with the `getConnected` handler?

mickmister:
> Mainly the issue is storing the token as part of user data in the KV store.

Storing the tokens in a separate KV entry seems like it would improve the setup of authentication, no matter what solution is used now or in the future. If we decide to do this, we also need to design a migration plan.

We could:
- perform a migration on plugin start-up
- perform an incremental migration (look for token and fall back on current storage location)
- require every user to reconnect their account

The last option is by far the easiest and safest method. It's inconvenient for admins/user to manage this, but all that needs to be done is to bump a major release of the plugin to signal the breaking change.

mickmister:
Given the strategy of storing the token separately, does this seem more feasible? The `getConnected` issue seems to be solved by this.

MatthewDorner:
Unless there was a "patch KV store JSON value" option (pretty sure there isn't) or if `getConnected` could re-load the userData after making its request and then apply the changes to that before saving it, though this could have the same problem if another request was running concurrently, and maybe other places in the code that modify user data could have the same problem with concurrency.

But storing the token separately should solve the issue, and of course I'd choose the one with no migration to be coded, but that depends on if you think that's acceptable to users.

Another thing I considered is taking this to GitLab and asking for a grace period to be implemented in the refresh token rotation which would make all this unnecessary, but I assumed it would take a long time for them to do it. If all we need to do is make the token stored separately, maybe that is not that much work though.

mickmister:
> But storing the token separately should solve the issue, and of course I'd choose the one with no migration to be coded, but that depends on if you think that's acceptable to users.

The plugin isn't usable with newer versions of GitLab right now, so this sort of breaking change is warranted IMO.

> Another thing I considered is taking this to GitLab and asking for a grace period to be implemented in the refresh token rotation which would make all this unnecessary, but I assumed it would take a long time for them to do it.

I wouldn't count on this, certainly not for the purpose of this PR.

> If all we need to do is make the token stored separately, maybe that is not that much work though.

Yeah I think storing the token with the user info was a design mistake. We couldn't have foreseen the sort of issues we're running into now. Storing the token separate makes the most sense to me. Again, all we need to do is bump the version to a new major version (2.0.0 in this case) for admins to see a notice to check the release notes. This will be done in a separate release bump PR.

mickmister:
Hey @MatthewDorner, what are your thoughts on moving the tokens to its own place in the kv store, and/or the other design decisions laid out above? I know this increases the scope quite a bit so I'd like to know how you feel about this direction.

MatthewDorner:
@mickmister I am finishing up implementing and will commit today or tomorrow. So far, it's more code but it's much cleaner in terms of how everything is organized. I can contain all the logic relating to loading token, refreshing token and checking and disconnecting the user in case of an invalid token inside of the `useGitlabClient` method.

Still, it is a lot of code and hard to read. This:

```
var unreads []*gitlabLib.Todo
err := p.useGitlabClient(user, func(info *gitlab.UserInfo, token *oauth2.Token) error {
	resp, err := p.GitlabClient.GetUnreads(ctx, info, token)
	if err != nil {
		return err
	}
	unreads = resp
	return nil
})
if err != nil {
	return err
}
```

replaces this:

```
unreads, err := p.GitlabClient.GetUnreads(ctx, user)
if err != nil {
	return err
}
```

And this happens at every point in `main` where `p.GitlabClient` is used.

MatthewDorner:
Storing the token separately also adds some code but I think it is necessary as discussed above. The way I have it now, the token is loaded inside of `useGitlabClient`, so the `gitlab` client methods receive a valid token and user info separately, and the `main` methods just pass the user info to `useGitlabClient` which uses it to load the token and refresh if necessary before calling the client.

mickmister:
> Still, it is a lot of code and hard to read. This:
> 
> And this happens at every point in `main` where `p.GitlabClient` is used.

Personally I think the code reads better this way. Even though the code is a bit verbose now, it's clear what is happening and seems solid to me. LGTM :+1:

MatthewDorner:
I still have to do more testing since I changed a lot in the token refreshing and mutex code but you can see the basic solution. I'll probably test today and request review if everything is good.

server/plugin.go
```diff
@@ -41,10 +43,13 @@ const (
...
 
 	chimeraGitLabAppIdentifier = "plugin-gitlab"
+
+	invalidTokenError = "401 {error: invalid_token}" //#nosec G101 -- False positive
 )
 
 var (
-	manifest model.Manifest = root.Manifest
+	manifest   model.Manifest = root.Manifest
+	tokenMutex sync.Mutex
```
mickmister, line 52: Note that we'll want to use `cluster.Mutex` instead of this mutex

server/plugin.go
```diff
@@ -29,7 +30,8 @@ import (
 )
 
 const (
-	GitlabTokenKey       = "_gitlabtoken"
+	GitlabUserInfoKey    = "_userinfo"
+	GitlabUserTokenKey   = "_usertoken"
```
mickmister, line 34: I think changing these key names is the right move since we are going to force users to reconnect anyway. This results in persisting stale `UserInfo/Token` structs in their old entries forever, but it seems more appropriate than having the wrong data type stored in `_gitlabtoken` or a similar situation

server/plugin.go
```diff
@@ -582,7 +645,15 @@ func (p *Plugin) HasProjectHook(ctx context.Context, user *gitlab.UserInfo, name
 // HasGroupHook checks if the subscribed GitLab Group has a webhook
 // with a URL that matches the Mattermost Site URL.
 func (p *Plugin) HasGroupHook(ctx context.Context, user *gitlab.UserInfo, namespace string) (bool, error) {
-	hooks, err := p.GitlabClient.GetGroupHooks(ctx, user, namespace)
+	var hooks []*gitlab.WebhookInfo
+	err := p.useGitlabClient(user, func(info *gitlab.UserInfo, token *oauth2.Token) error {
```
mickmister, line 649: Since `useGitlabClient` doesn't actually provide a client, maybe we should name it `useGitlabToken`?

server/plugin.go
```diff
@@ -599,22 +670,84 @@ func (p *Plugin) HasGroupHook(ctx context.Context, user *gitlab.UserInfo, namesp
...
-			return nil, errors.Wrap(err, "unable to get the new refreshed token")
+func (p *Plugin) refreshToken(userInfo *gitlab.UserInfo, token *oauth2.Token) (*oauth2.Token, error) {
+	conf := p.getOAuthConfig()
+	src := conf.TokenSource(context.Background(), token)
+
+	newToken, err := src.Token() // this actually goes and renews the tokens
+
+	if err != nil {
+		if strings.Contains(err.Error(), "\"error\":\"invalid_grant\"") {
+			p.handleRevokedToken(userInfo)
```
mickmister, line 681: Just so I understand, this will only run when we have a token stored, and the token was revoked externally. In this case, the mutex will correctly only allow this to run once.

server/plugin.go
```diff
@@ -599,22 +670,84 @@ func (p *Plugin) HasGroupHook(ctx context.Context, user *gitlab.UserInfo, namesp
...
 		}
-		if newToken.AccessToken != token.AccessToken {
-			return newToken, nil
+		return nil, errors.Wrap(err, "unable to get the new refreshed token")
+	}
+
+	if newToken.AccessToken != token.AccessToken {
+		p.API.LogDebug("Gitlab token refreshed.", "UserID", userInfo.UserID, "Gitlab Username", userInfo.GitlabUsername)
```
mickmister, line 687: We want to avoid logging user-identifying information like usernames to the logs

```suggestion
		p.API.LogDebug("Gitlab token refreshed.", "UserID", userInfo.UserID)
```

Also I think this may be more spammy than useful since this will be logged per user every hour. @hanzei What are your thoughts on having this log message here?

mickmister: Also, should we also be comparing `RefreshToken` to check if we need to store the token? Not sure if it's possible for that to change but not the access token

server/plugin.go
```diff
@@ -599,22 +670,84 @@ func (p *Plugin) HasGroupHook(ctx context.Context, user *gitlab.UserInfo, namesp
...
+		newToken.AccessToken = unencryptedNewToken
+		return newToken, nil
+	}
+
+	return token, nil
+}
+
+func (p *Plugin) handleRevokedToken(info *gitlab.UserInfo) {
+	p.disconnectGitlabAccount(info.UserID)
+	err := p.CreateBotDMPost(info.UserID, "Your GitLab account was disconnected due to an invalid or revoked authorization token. Reconnect your account using the `/gitlab connect` command.", "custom_git_revoked_token")
```
mickmister, line 703: As we discussed offline, maybe we should check if the disconnect was successful before sending this message. This way we don't spam the user with these messages on concurrent attempts to access the API.

@hanzei Note that this only occurs when the token was literally revoked externally, e.g. from Gitlab's UI

server/webhook.go
```diff
@@ -206,22 +216,47 @@ func (p *Plugin) permissionToProject(ctx context.Context, userID, namespace, pro
...
+			return nil
+		})
+		if getProjectErr != nil {
+			return nil, getProjectErr
 		}
-		newWebhook, err := gitlabClient.NewProjectHook(ctx, info, project.ID, hookOptions)
-		if err != nil {
-			return nil, err
+		var newWebhook *gitlab.WebhookInfo
+		getGroupErr := p.useGitlabClient(info, func(info *gitlab.UserInfo, token *oauth2.Token) error {
```
mickmister, line 235: Should we call `useGitlabClient` only once here for both operations, to avoid re-fetching the token?

mickmister (changes requested):
Great work on this @matthewdorner. Looking forward to having this merged and released :+1:

In general this is looking awesome. On this iteration, I have a few comments/questions, and a request to use `cluster.Mutex` instead of `sync.Mutex`

Also something comes to mind, resolving merge conflicts with parallel efforts on the project is going to be a pain. Hopefully that's not going to be too complicated for whoever does the conflict resolutions.

server/plugin.go
MatthewDorner: I figured the regular mutex would perform better (maybe ? because the cluster one uses KV) and possibly be more stable. Do we really need the mutex to be user specific if it's only happening during the refresh? Or do we still need the cluster one to support cluster instances?

server/plugin.go
MatthewDorner: This will only run when the token was revoked externally, and the token expiration time is within one minute of the request (since this is in `p.refreshToken`, it occurs when we try to refresh).

If the token is not marked to expire, we'll try to use it and it will fail with the `401 {error: invalid_token}` error (a different error when we try to make a request through the `go-gitlab` library). And if it fails there, we call `p.handleRevokedToken` from `useGitlabClient` so the result is the same.

server/plugin.go
```diff
@@ -225,68 +230,84 @@ func (p *Plugin) getOAuthConfigForChimeraApp(scopes []string, redirectURL string
...
 	config := p.getConfiguration()
 
-	encryptedToken, err := encrypt([]byte(config.EncryptionKey), info.Token.AccessToken)
+	encryptedToken, err := encrypt([]byte(config.EncryptionKey), token.AccessToken)
 	if err != nil {
 		return err
 	}
 
-	info.Token.AccessToken = encryptedToken
+	token.AccessToken = encryptedToken
```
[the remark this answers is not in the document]
MatthewDorner: Yes, definitely

server/plugin.go
MatthewDorner: It could be either way. The name made sense to me since we're using it to wrap the calls to `p.GitlabClient` methods.

server/webhook.go
MatthewDorner: I suppose we could, though it's only happening when a webhook is created, so the performance impact would be minor. The better case would be if it's more readable.

server/plugin.go
MatthewDorner: And making this change would require another use of a mutex, correct?

server/plugin.go
MatthewDorner: This is all left over from the previous token refresh code before I started working on the issue. The `RefreshToken` should not change unless the `AccessToken` changes, though in real use we always receive a different token so the entire check could be unnecessary.

server/plugin.go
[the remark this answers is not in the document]
MatthewDorner: Alright, I just committed this. Now, it's harder to test since you can't edit the expiration in the database. You can instead test by adding `tok.Expiry = time.Now()` in `completeConnectUserToGitlab` before the new token is saved. Then, the token refresh will be triggered by the app as it refreshes at the end of the connect user operation.

server/plugin.go
MatthewDorner: Committed your suggested change to remove the GitLab username.

server/plugin.go
MatthewDorner: Committed the change to use `cluster.Mutex`

DHaussermann:
@mickmister or @MatthewDorner can you confirm next steps on this? We will need a GitLab release to go along with `v7.6.0` and if this is solved we need the solution merged. 

cc @asatkinson

MatthewDorner:
@DHaussermann From my perspective, it is waiting for review and requested changes have been addressed. Happy to make any further changes if necessary.

MatthewDorner:
I'll get those merge conflicts fixed, looks like they just appeared.

server/plugin.go
MatthewDorner: @mickmister Haven't implemented this change, waiting on reviewers' decision on if it's necessary.

server/webhook.go
MatthewDorner: @mickmister Haven't implemented this, felt it's sufficient to leave as is unless reviewers feel otherwise.

server/plugin.go
```diff
@@ -635,22 +706,87 @@ func (p *Plugin) HasGroupHook(ctx context.Context, user *gitlab.UserInfo, namesp
...
-		if err != nil {
-			return nil, errors.Wrap(err, "unable to get the new refreshed token")
+func (p *Plugin) refreshToken(userInfo *gitlab.UserInfo, token *oauth2.Token) (*oauth2.Token, error) {
+	conf := p.getOAuthConfig()
+	src := conf.TokenSource(context.Background(), token)
+
+	newToken, err := src.Token() // this actually goes and renews the tokens
+
+	if err != nil {
+		if strings.Contains(err.Error(), "\"error\":\"invalid_grant\"") {
```
hanzei, line 716: Nit: Should the error message be a global const?

server/plugin.go
hanzei: Surely, `disconnectGitlabAccount` could return an `error` (maybe even a multierror), but this seems like a nit pick to me.

server/plugin.go
hanzei: 1/5 consistency is key. If we ever want to add custom post rendering, the post are already marked.

hanzei (changes requested):
First of all, thank you for all your work on this PR @MatthewDorner. You did a fantastic job investigating the issue and working on a proper solution. :+1: 

In general, I quite like the approach that the PR is taking. The locking around the token refreshing looks good and the UX improvement is great.

The use of `useGitlabClient` looks cluncy to me, but that is a style issue that can be resolved later. IMO the gitlab client should take care of the token refresh.

The only concern from my side the is lack of a migration for the OAuth token. Could we do a lazy migration here i.e. load the token from store and if it's not found, load the UserInfo and use the token from there?

server/plugin.go
```diff
@@ -635,22 +706,87 @@ func (p *Plugin) HasGroupHook(ctx context.Context, user *gitlab.UserInfo, namesp
...
+			p.handleRevokedToken(userInfo)
 		}
-		if newToken.AccessToken != token.AccessToken {
-			return newToken, nil
+		return nil, errors.Wrap(err, "unable to get the new refreshed token")
+	}
+
+	if newToken.AccessToken != token.AccessToken {
+		p.API.LogDebug("Gitlab token refreshed.", "UserID", userInfo.UserID)
```
mickmister, line 723: This may get logged too often. I'm thinking we should remove this. @hanzei What do you think?

mickmister (approved):
Thanks for your work on this and your patience on the reviews @MatthewDorner

Overall LGTM :+1:

Just one question regarding logging. +1 to @hanzei's comments above

server/plugin.go
hanzei: I plan to add a trace logger that only log to debug when a specific setting is activated. This would be a good log message for the trace logger.

Let's add a comment to do that later.

server/plugin.go
MatthewDorner: I just left it that way since it didn't return an error to begin with and it didn't seem particularly relevant to this PR to change that behavior.

MatthewDorner:
@hanzei
> Could we do a lazy migration here i.e. load the token from store and if it's not found, load the UserInfo and use the token from there?

Sounds OK to me and easy to implement. @mickmister agree?

mickmister:
@MatthewDorner Yes this sounds good to me :+1:

server/plugin.go
MatthewDorner: I didn't because it's only used in one place, but it's a matter of your preferred conventions.

MatthewDorner:
@hanzei @mickmister 
Implementing the migration. I think the migration needs to be locked under the same mutex as the token refresh. Otherwise, a migration could overwrite a token that's just been refreshed. Also to avoid any issue with multiple migrations running concurrently, though that's probably less of a problem.

In order to do this with the `cluster.Mutex`, can I simply call `cluster.NewMutex(p.API, info.UserID+"-refresh-token")` again in the migration method to essentially get the same mutex, or do I need to make the mutex globally accessible? I'm guessing the former.

MatthewDorner:
Did a first write of the KV migration, which seems to work. I'll clean up, continue testing and fix merge conflicts.

server/plugin.go
```diff
@@ -280,15 +282,76 @@ func (p *Plugin) deleteGitlabUserToken(userID string) error {
...
 	return &userInfo, nil
 }
 
+func (p *Plugin) migrateGitlabToken(userID string) (*gitlab.UserInfo, *APIErrorResponse) {
+	config := p.getConfiguration()
+
+	var userInfo struct {
+		gitlab.UserInfo
+		Token *oauth2.Token
+	}
```
mickmister, line 307: Just so I understand correctly, we need to declare this local struct, because we no longer use that token from this struct elsewhere, so we have removed it from the "known" struct definition in the KV store.

server/plugin.go
```diff
@@ -280,15 +282,76 @@ func (p *Plugin) deleteGitlabUserToken(userID string) error {
 func (p *Plugin) getGitlabUserInfoByMattermostID(userID string) (*gitlab.UserInfo, *APIErrorResponse) {
 	var userInfo gitlab.UserInfo
 
-	if infoBytes, err := p.API.KVGet(userID + GitlabUserInfoKey); err != nil || infoBytes == nil {
-		return nil, &APIErrorResponse{ID: APIErrorIDNotConnected, Message: "Must connect user account to GitLab first.", StatusCode: http.StatusBadRequest}
+	infoBytes, err := p.API.KVGet(userID + GitlabUserInfoKey)
+
+	if err != nil || infoBytes == nil {
+		gitlabTokenBytes, err := p.API.KVGet(userID + GitlabMigrationTokenKey)
+		if err != nil || gitlabTokenBytes == nil {
```
mickmister, line 289: I feel that this logic should go in the place where we attempt to fetch the token, rather than where we fetch the user info. The token is what's relevant in this migration, so we would want to "fall back" on using the token on first use if necessary.

MatthewDorner: Not sure i understand, I'm doing the migration here because userinfo is the first thing we query, and we only deal with token later which is pretty tied to the decision made about how to solve the issue in this PR generally.

If the migration need to be done, the query from KV will fail. Unless I change it so the key is the same and conditionally unmarshal to a type with/without the token field, but that just seems much more complicated.

mickmister: I was saying that in the spot where we already using the mutex, we would do:

- Check if token exists in new KV location
- If not, check in old KV location. If it's in the old location, migrate to new location
- Use token whether we got it from old or new location

Is this how you understood my comment?

MatthewDorner: I think so. But in order to get to that point where we are already using the mutex, we have loaded the userinfo in a prior step (which is not locked under mutex, and `p.getGitlabUserInfoByMattermostID` is called from lots of different places in code depending on what event is being handled). But if the migration needs to be done, we will fail to load the userinfo from the KV at that point, which is why I put the migration there.

To me, this relates to the decision of placing the token handling logic close to the gitlab client (inside the callback, which can then be locked under mutex if necessary), and allowing the userinfo to be stored/retrieved freely anywhere without mutex which is why we needed to separate these pieces of data into two KV entries. But in this case, a mutex is still needed for the migration just because we are also migrating the token.

mickmister: > But if the migration needs to be done, we will fail to load the userinfo from the KV at that point, which is why I put the migration there.

I guess I don't understand this part. Why will we fail to load the userinfo in that case?

MatthewDorner: We will fail to load it because I changed the KV key and datatype that it is unmarshalled to. I see that I can try to do it as you're saying, where I guess I would keep the same KV key and just add a new one for the token, but I anticipate this might be messier in some ways. I can look into it.

MatthewDorner: I suppose if this makes it possible to avoid using the mutex in a second place in the code, that is preferable.

To clarify, do you imagine just keeping the old KV key for userdata and creating a new one for the token only, or are we still migrating to two new keys and deleting the old one?

MatthewDorner: Actually I don't think it will help with the mutex as needing a refresh and needing a migration are still completely separate conditions. Also if the old userdata + token are to be loaded without mutex lock and the migration is done only later when we attempt to fetch token from the new key, it's a possible issue that we are carrying the old token around that was loaded without mutex although I guess I'd re-load it with mutex when migrating as I'm doing in the migration now. But that's an example of how the current solution seems cleaner / safer to me: just immediately doing the migration.

mickmister: > I see that I can try to do it as you're saying, where I guess I would keep the same KV key and just add a new one for the token

Yeah I think we should do it this way. I don't think there's any reason to relocate the `UserInfo` data to a new key. It should stay in its current location, whereas the token is the exception to the rule, that it has to go somewhere else.

Also sorry for missing these posts. It's easier for me to stay up with conversations if I'm mentioned in the responses

> Actually I don't think it will help with the mutex as needing a refresh and needing a migration are still completely separate conditions.

Even though they're separate conditions, they touch the same data. I think they can be done in the same spot/mutex in the token usage block

> it's a possible issue that we are carrying the old token around that was loaded without mutex although I guess I'd re-load it with mutex when migrating as I'm doing in the migration now.  But that's an example of how the current solution seems cleaner / safer to me: just immediately doing the migration.

Yes I think we should re-load it with the mutex at the time of trying to use the token. I think having the migration at the spot of using the token would be cleaner, since we would just have one complicated spot where these store modifications will occur (migrate and token refresh)

mickmister: @hanzei What are you thoughts on the two strategies laid out above?

MatthewDorner: @mickmister
- The old KV key for both token and userinfo is `_gitlabtoken`. If the token is no longer stored there but we keep the same key, it's a bit confusing.
- Before adding the migration, mutex was only used if we detect the token is expired. I don't see how we can add the migration functionality into the same function in order to minimize the use of the mutex. If it's something like "if token is expired, OR if migration needs to be performed, then run this function that both refreshes if necessary and migrates if necessary" seems like bad design, and that we should just have another function (as I currently do) that also uses the mutex. Or maybe use a wrapper like `.withTokenMutex` and then these two separate functions? That aside, I can still see the point in locating both mutex operations closer to each other.

MatthewDorner:
@mickmister 
I moved the migration code, but so far left it so it's still migrating to two new KV keys and deleting the old one. It would simplify things, but it depends on whether you want to leave the old `_gitlabtoken` key now that it no longer stores the token there.

I also see that I will have conflicts with https://github.com/mattermost/mattermost-plugin-gitlab/pull/364. The fact that the new `p.API.KVGet` can handle the unmarshal as well should help to clean up some of the code in this PR.

server/plugin.go
mickmister: > I didn't because it's only used in one place, but it's a matter of your preferred conventions.

When a constant has a significant meaning like this (essentially an error id in the external system), we generally put it at the top of the file to show that it is truly a constant

mickmister (approved):
LGTM, thanks so much @MatthewDorner :+1:

DHaussermann:
Hi @MatthewDorner I have started testing by looking at users who never had an invalid token to see how this behaves. I tried on 2 severs. I connected the users and waited. On 1 server I see no issues but on my local I'm seeing a problem where somehow the token became invalid. 

Note that this happened about a day and a half later
image
I don't believe I made any changes.    

Sorry do to an unrelated issue on my server I can't get corresponding logs.

I will now test this again from a clean state with logging working and post an update. I was just curious if you knew of a typical scenarios where this may occur.

MatthewDorner:
@DHaussermann that should only happen due to something on the GitLab side. Are you using cloud GitLab or your own? What version of GitLab and are you sure nothing changed during that time?

DHaussermann:
@MatthewDorner I have not been able to reproduce any unexpected behavoior. I'm curious if what I saw yesterday was caused by the oAuth app on the GitLab side being revoked. (As I think is one of you possible suggestions above.)

For Now I have the build on multiple servers to see if anything unexpected happens.  

So far as testing the solution when the token is invalid - I am not able the invalidate an individual users token in the KV store. That said, I am able to invalidate the token for all users by revoke access the the entire oAuth App as a loose equivalent. When doing so, I see that on upgrade users with invalid tokens receive the DM notification as expected. I also see that users with valid token do not get the DM and their user mapping remains functional 👍 

I will give this 1 more half day and explore a bit more including confirming the logging behavior when users are in an invalid state. 

@mickmister just a heads up - after this if no issues are found, it may be far more efficient to see if we can upgrade the community server (as we can see the core concept working in test) and then confirm on community if individual users get the notifications as expected.

mickmister:
@DHaussermann Sounds good to me :+1:

@MatthewDorner It looks like there are more conflicts to resolve. LMK if you'd like me to do this since you've spent an immense amount of time on this already

DHaussermann:
@MatthewDorner I think I've done all the testing I can before we get this deployed to Community. In Summary:
- Spamming the logs when trying to fetch data from the API stops on upgrade as the plugin does not treat the affected users as connected
- Affected users get the DM notice to reconnect
- Users with a functional authentication continue to us the plugin normally  
- After 2 days of monitoring I see nothing unusual in the logs after the user has re-connected
- Regression tested Chimera proxy
- Regression tested authentication for browser and desktop
 
There is one thing I wanted to raise here. I sometimes see the DM message about invalid tokens show 3 or 4 times. 
As the message occurs the next time the plugin tries to fetch the data and then realizes the user is no longer authenticated and there are 4 values, possibly its posting 1 message for each?
Any thoughts on this or if it would be easy to solve?
image

mickmister:
> There is one thing I wanted to raise here. I sometimes see the DM message about invalid tokens show 3 or 4 times.
> As the message occurs the next time the plugin tries to fetch the data and then realizes the user is no longer authenticated and there are 4 values, possibly its posting 1 message for each?

@DHaussermann Do you have exact repro steps for this? For example - If the user refreshes, do they receive another 4 notifications?

This duplicate DM behavior is one of the main things the PR tries to avoid, so that's unfortunate that this is happening

MatthewDorner:
@mickmister @DHaussermann This is happening due to `handleRevokedToken` getting called from `useGitlabClient` AFTER the client call is made. Because when the authorization is revoked on the GitLab side (which should be the only case where this occurs), we don't know until after we attempt to make the call. All of the mutex concurrency control I implemented happens BEFORE the client call, in order to ensure that we have a valid token as far as we can tell and that if we need to refresh the token, concurrent requests will wait until they are able to use the new token.

So this means fixing the multiple DM issue would require another mutex use. Unless we could somehow just check in `disconnectGitlabAccount` if the keys are already deleted, and return an error if so, so that the `handleRevokedToken` could not send the DM if `disconnectGitlabAccount` returns error. But I think this is still flaky with concurrency and would require a mutex anyway.

@mickmister note that the multiple DM issue is separate from the "spamming the error log" issue, where requests were failing should be fixed according to the above
> Spamming the logs when trying to fetch data from the API stops on upgrade as the plugin does not treat the affected users as connected

MatthewDorner:
@DHaussermann also I should note, the spamming error log issue (if I'm thinking of the same thing that happens every 2 hours) is not related to users being connected or not. It's because the 4 concurrent requests are happening when the token needs to refresh (every 2 hours) and without concurrency control 3 of these requests will fail because they try to use the old token after one request has refreshed it.

As I understand it the issue this PR is for was created to begin to address this issue by doing better error handling for those cases (disconnect the user) but the PR grew to discover and address the root cause as well which is why it got so complicated.

DHaussermann:
@MatthewDorner and @mickmister 
0/5 Is it possible that this may not be worth the technical complexity to solve? To my understanding the muttiple DMs would happen only once correct?

cc @asatkinson

mickmister:
> 0/5 Is it possible that this may not be worth the technical complexity to solve? To my understanding the muttiple DMs would happen only once correct?

@DHaussermann If we are okay with the multiple DM behavior on a product level, I would say let's get this merged and address that separately

mickmister:
A potential solution is to avoid the 4 DMs is to make the API endpoint handler for `getConnected` contact GitLab's server to assert that the token is still valid, rather than just checking the KV store. The frontend always call `getConnected` before calling those 4 LHS count methods. I think the point of `getConnected` is to confirm that the user is connected, so it may make sense to fully verify that.

srkgupta (approved):

mickmister:
@srkgupta Do you mind giving this another review? I resolved some merge conflicts regarding the `p.API -> p.client` migration

@DHaussermann Are you able to give this another smoke test?

DHaussermann (approved):
Tested and passed
Sanity checked the functionality now that @mickmister has resolved merge conflicts.
- Requests for GitLab dada stop when token is known to be invalid
- DMs are receive 
- Reconnect works and user remains conneted

LGTM!

srkgupta (approved):

mickmister:
Looks like we need @hanzei's approval before this can be merged:

> Merging can be performed automatically once the requested changes are addressed.

PR #348, opened 2022-12-15
Avoid kvset/kvget race on subscriptions update

lieut-data:
#### Summary
When adding or deleting a subscription, we save the subscriptions via `KVSet`, and previously turned around to fetch those with `KVGet` before broadcasting to the client. On installations with a database replica, the second request might not hit the `master` database, and the update just effected might not be "seen".

Skip the `KVGet` and instead just use the information about the updated susbcriptions that we already have in hand when updating the client.

#### Ticket Link
Fixes: https://github.com/mattermost/mattermost-plugin-gitlab/issues/342

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -482,7 +482,7 @@ func parseTriggers(triggersCsv string) *gitlab.AddWebhookOptions {
 
 func (p *Plugin) subscriptionDelete(info *gitlab.UserInfo, config *configuration, fullPath, channelID string) (string, error) {
 	normalizedPath := normalizePath(fullPath, config.GitlabURL)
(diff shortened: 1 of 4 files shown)


mickmister (approved):

spirosoik (approved):

hanzei:
@DHaussermann The PR is ready for your review

DHaussermann (approved):
Tested and passed
- Found an affected server and upgraded to ensure the issue is resolved
- Delete command is always reflected in the RHS in real-time
- Tested on in HA as well 
No issues found 

LGTM!

PR #363, opened 2023-03-09
[MM-51269] Log errors from telemetry events

hanzei:
#### Summary
Previously errors returned from the telemetry tracker didn't get logged. This PR adds proper log messages for all errors the telemetry tracker returns.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-51269


m1lt0n (commented):
Looks good. Question: What about the `TrackUserEvent` calls from `flow` that are directly using the tracker's `TrackUserEvent`?

mickmister (approved):

PR #364, opened 2023-03-13
Remove usage of model.AppError

srkgupta:
#### Summary
Remove usage of model.AppError in Gitlab plugin

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50986


mickmister (approved):
LGTM :+1:

m1lt0n (approved):
Looks good! Only comment: there is an occurrence of `model.AppError` in `command.go` in the `ExecuteCommand`, is this possible to be updated too or not?

srkgupta:
> Looks good! Only comment: there is an occurrence of model.AppError in command.go in the ExecuteCommand, is this possible to be updated too or not?

Thanks @m1lt0n. Nice catch. I have updated the occurrence in ExecuteCommand function too.

server/plugin.go
```diff
@@ -259,16 +259,16 @@ func (p *Plugin) getGitlabUserInfoByMattermostID(userID string) (*gitlab.UserInf
...
 
 	var userInfo gitlab.UserInfo
-
-	if infoBytes, err := p.API.KVGet(userID + GitlabTokenKey); err != nil || infoBytes == nil {
+	var infoBytes []byte
+	if err := p.client.KV.Get(userID+GitlabTokenKey, &infoBytes); err != nil || infoBytes == nil {
 		return nil, &APIErrorResponse{ID: APIErrorIDNotConnected, Message: "Must connect user account to GitLab first.", StatusCode: http.StatusBadRequest}
 	} else if err := json.Unmarshal(infoBytes, &userInfo); err != nil {
 		return nil, &APIErrorResponse{ID: "", Message: "Unable to parse token.", StatusCode: http.StatusInternalServerError}
 	}
```
hanzei, line 267: `p.client.KV.Get` can also take of unmarshling the json payload. Let's take advantage here to simplify the code.

server/subscriptions.go
```diff
@@ -92,9 +92,10 @@ func (p *Plugin) AddSubscription(fullPath string, sub *subscription.Subscription
 func (p *Plugin) GetSubscriptions() (*Subscriptions, error) {
 	var subscriptions *Subscriptions
 
-	value, err := p.API.KVGet(SubscriptionsKey)
+	var value []byte
+	err := p.client.KV.Get(SubscriptionsKey, &value)
```
hanzei, line 96: Same here

hanzei (changes requested):
Nice work, @srkgupta :+1: 

I left a few comments below. The most important item is to check if the slash command still works.

hanzei:
Nit pick: https://github.com/mattermost/mattermost-plugin-gitlab/blob/c7080b6e085e2133b84ad08f2126387a50563590/server/command.go#L95 can be updated to `iconData, err := command.GetIconData(&p.client.System, "assets/icon.svg")` so `p.API` is used even less,

srkgupta:
Hi @hanzei 

I have made all the changes based on your feedback. Requesting your re-review on this.

Thanks

server/plugin.go
```diff
@@ -77,8 +77,6 @@ type Plugin struct {
 }
 
 func (p *Plugin) OnActivate() error {
-	p.client = pluginapi.NewClient(p.API, p.Driver)
```
hanzei, line 80: I'm not quite sure if it's really needed, but should we add a 
```golang
	if p.client == nil {
		p.client = pluginapi.NewClient(p.API, p.Driver)
	}
```
just for additional protection against races?

hanzei (approved):
LGTM :rocket:

server/plugin.go
srkgupta: Sure. Handled the race condition. Thanks.

hanzei:
@DHaussermann Heads up that this PR is ready for your review

DHaussermann (approved):
Tested and passed
- Setup and smoke tested GitLab
- Connect and disconnect from browser and desktop
- Various commands to retrieve data
- Created subscription and ensured deliveries are working
- RHS works to show connection and subscribe options
- Lists of work items open in RHS as expected  
- No changes in behavior found
- Nothing unusual in the logs

LGTM!

PR #393, opened 2023-08-21
[MM-53944] Make use of github.com/mattermost/mattermost/server/public

hanzei:
#### Summary
This PR makes use of the newly created go module in https://pkg.go.dev/github.com/mattermost/mattermost/server/public.

See https://github.com/mattermost/mattermost-plugin-demo/pull/163 for a similar migration.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53944


mickmister (approved):
LGTM

hanzei:
@spirosoik Would you mind giving this PR a review so we can get it into 1.7.0?

spirosoik (commented):
@hanzei overall looks good to me but I don't know what could be the impact of this change. may we need to include one more pair of eyes here. what do you think?

hanzei:
I did a smoke test of the change (compile the plugin, start it, and run a slash command). Is that enough? More complex issues might also be caught during release testing.

PR #394, opened 2023-08-22
[GH-271] Updated logic for sidebar header to show assigned MRs instead of opened MRs

raghavaggarwal2308:
#### Summary
Changes:
1. Get assigned MRs instead of the user's MRs.
2. Updated the name in the tooltip of the sidebar buttons.
3. Updated API paths.
4. Updated variable and function names.
5. Updated icons in the sidebar.
6. Updated documentation.

#### Screenshots
Before:
image

After:
image

#### Ticket Link
Fixes #271

PR #438, opened 2024-01-11
[GH-394] Migration to TypeScript: Sidebar_button migrated to Typescript

Willy-Wakam:
#### Summary

- The file sidebar_button.jsx was migrated into his appropriated Typescript version

- All properties and classes were well typed 

#### Ticket Link
Fixes [https://github.com/mattermost/mattermost-plugin-gitlab/issues/430](https://github.com/mattermost/mattermost-plugin-gitlab/issues/430)


webapp/src/components/sidebar_buttons/sidebar_buttons.tsx
```diff
@@ -1,36 +1,45 @@
...
+import { Placement } from 'react-bootstrap/esm/Overlay';
+
+interface SidebarButtonsProps {
+    theme: any;
+    connected: boolean;
+    username: string;
+    org: string;
+    clientId: string;
+    gitlabURL: string;
+    reviews: any[];
```
mickmister, line 19: One thing we want to avoid is putting `any` on the types, as this can sort of defeat the purpose of using typescript. We're only checking the length of the arrays in this file, so we don't necessarily need the type. If you would like to type them from these types you can. Otherwise we can just use `unknown` instead of `any`

```suggestion
    reviews: unknown[];
```

webapp/src/components/sidebar_buttons/sidebar_buttons.tsx
```diff
@@ -1,36 +1,45 @@
...
+}
 
-import {GitLabIssuesIcon, GitLabMergeRequestIcon, GitLabReviewsIcon, GitLabTodosIcon} from './button_icons';
+interface SidebarButtonsState {
+    refreshing: boolean;
+}
 
-export default class SidebarButtons extends React.PureComponent {
+export default class SidebarButtons extends PureComponent<SidebarButtonsProps, SidebarButtonsState> {
     static propTypes = {
```
mickmister, line 38: We can remove this, and instead use https://www.npmjs.com/package/babel-plugin-typescript-to-proptypes, and add the plugin to the babel config file. This automatically adds the propTypes for us

mickmister (commented):
Thanks for this contribution @Willy-Wakam! This is looking really good. I have just a few suggestions. Also, we'll want to convert `src/components/sidebar_buttons/index.js` to typescript if possible.

To fix some linting rules, I suggest changing these lines in `.eslintrc.json`:

```
"no-magic-numbers": "off",
"no-use-before-define": "off",
```

Note that `no-use-before-define` is superseded by `@typescript-eslint/no-use-before-define`, so those checks will still be made.

Can you please run this command locally to ensure the linting is correct? The `--fix` flag makes it automatically fix certain things as well.

```sh
./node_modules/.bin/eslint src/components/sidebar_buttons/sidebar_buttons.tsx --fix
```

I tried to run the below command locally but it doesn't seem to work. It spits out a bunch of errors from `node_modules/@types` only. Using your editor's error reporting in that file should be enough for tsc errors.

```sh
npx tsc src/components/sidebar_buttons/sidebar_buttons.tsx --noEmit --skipLibCheck
```

mickmister:
@ayusht2810 Heads up that the linter seems to be configured incorrectly in `npm run lint`, as we are ignoring `js` and `jsx` files for some reason. That must have been done when we first introduced linting for some reason

webapp/src/components/sidebar_buttons/sidebar_buttons.tsx
Willy-Wakam: Could you please provide me an example with the possible solution, because I've tried, but I still get some errors when removing the interface and adding them in the propTypes

mickmister: We should just need to do this:

```
npm i -D babel-plugin-typescript-to-proptypes
```

Then add the babel plugin like so:
https://github.com/mattermost/mattermost-plugin-jira/blob/a4d725ccc4e898d58ebf3bb054f72f7cb30fefc6/webapp/babel.config.js#L31

Then we can remove the `static propTypes` thing altogether in the typescript file, since those will automatically be generated

Willy-Wakam: After running the command, I got this error 
error

And the babel plugin was already configured in the babel.config-file

ayusht2810 (approved):
LGTM! Apart from @mickmister comments.

webapp/package-lock.json
ayusht2810, line 1: why are we having changes in this file even though we didn't make any changes in the `package.json` file?

Willy-Wakam: I got some errors in the previous commit, so I deleted and pulled the forked project again. I guess by installing node modules, the file was edited, although it was not really changed

Willy-Wakam:
If there are no more relevant changes, could someone merge and close the PR?

mickmister (approved):
LGTM, thanks @Willy-Wakam!

webapp/package-lock.json
mickmister: The `package-lock.json` says it's `lockfileVersion: 3` now. @Willy-Wakam Can you please either revert those changes, or try re-running `npm install` in the `webapp` folder after setting your node version to the version defined here?

https://github.com/mattermost/mattermost-plugin-gitlab/blob/eab819326c87ebf92d440d591e1268f69cfb60fc/.nvmrc#L1

You can set your node version using tools like https://github.com/nvm-sh/nvm or https://github.com/Schniz/fnm

ayusht2810 (approved):
LGTM! Apart from lock file changes.

mickmister:
Thanks for the contribution @Willy-Wakam!

PR #577, opened 2025-05-30
Upgrade mattermost-redux to 10.8.0

abbas-dependable-naqvi:
#### Summary
Upgrade mattermost-redux to 10.8.0


webapp/package-lock.json
hmhealey, line 1: Just to note, a lot of these lines changed are due to the lockfile format updating from 2 to 3. We've used version 3 elsewhere for a few years now, so it's a good change

hmhealey (changes requested):
For those broken type imports, the Theme type is now in `mattermost-redux/selectors/entities/preferences`

webapp/package-lock.json
abbas-dependable-naqvi: @wiggin77 Please share the screenshot for the snyk bot issue.

webapp/package.json
```diff
@@ -83,5 +84,8 @@
         "typescript": "5.6.3",
         "webpack": "5.96.1",
         "webpack-cli": "5.1.4"
+    },
+    "overrides": {
```
abbas-dependable-naqvi: @hmhealey Can't get this working without override. Will this do or we should fix it?

hmhealey: Looks like we need to fix something upstream for this to work without the override.

Could you change that line to `"@mattermost/types": "$@mattermost/types"` though? That'll set the override to use the version number defined above so that we don't have to update the override when we update this in the future

hmhealey (changes requested):

hmhealey (approved):

PR #595, opened 2025-07-30
Save multiple instances in KV Store

abbas-dependable-naqvi:
#### Summary
Save multiple instances in KV Store

### DIFF
--- a/server/utils.go
+++ b/server/utils.go
@@ -299,3 +299,22 @@ func lastN(s string, n int) string {
 
 	return string(out)
 }
+
+func containsString(slice []string, value string) bool {
+	for _, v := range slice {
+		if v == value {
+			return true
+		}
+	}
+	return false
+}
+
+func removeStringFromSlice(list []string, target string) []string {
+	var updated []string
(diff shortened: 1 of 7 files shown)


abbas-dependable-naqvi:
Fixed

server/utils.go
```diff
@@ -299,3 +299,23 @@ func lastN(s string, n int) string {
...
+func containsString(slice []string, value string) bool {
+	for _, v := range slice {
+		if v == value {
+			return true
+		}
+	}
+	return false
+}
+
+//nolint:unused
```
wiggin77, line 312: Why is the nolint needed?

wiggin77 (approved):

server/utils.go
abbas-dependable-naqvi: The code for instance deletion was added along the other CRUD functions but the actual functionality is added in the next PR

Yash-Chakerverti (approved):
I have done the sanity of this PR and all the functioanlities are working as expected. LGTM, Approved

PR #612, opened 2025-09-29
fix: properly retrieve all users that have access to a project

buzzyboy:
#### Summary
This fixes an issue related to the available assignees when using the `/gitlab issue create` command. Currently, it only shows you direct project members; this change updates the Gitlab SDK to use `ListAllProjectMembers` instead so that it gets all users with inherited membership to the group.

I'm not sure what downwind effects this will have with other functions that call this, but this would appear to at least fix my issue related to the ticket I linked.

### DIFF
--- a/server/gitlab/api.go
+++ b/server/gitlab/api.go
@@ -786,7 +786,7 @@ func (g *gitlab) GetProjectMembers(ctx context.Context, user *UserInfo, projectI
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


raghavaggarwal2308:
> For me this change looks correct if we want to show all members that have access to a given project, but I'm not sure if that method was used by any particular reason. The PR that introduced this was #306

@fmartingr I dont think there was any specific reason for that, we can keep this change

PR #663, opened 2026-03-30
MM-68133 Removing default instance setting on connect

avasconcelos114:
#### Summary

This PR de-couples the setting of a default instance from the connection flow so that it remains a part of instance setup or something used through its own dedicated command

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-68133

## Change Impact: 🟡 Medium

**Reasoning:** The PR decouples default instance setting from the connect flow and adds authorization checks to the setup flow. While this is a clean architectural improvement, it spans multiple flow management entry points and touches authorization logic in areas affecting user onboarding.

**Regression Risk:** 
- Low risk for the `handleConnect` simplification: removing the auto-setting of default instance is a straightforward change that reduces side effects and makes the connection flow more explicit.
- New admin authorization checks in `stepSetDefaultInstance` could impact users in the setup/oauth flows if they attempt to set a default instance but lack system admin privileges. However, this is an intentional security enhancement and there's a fallback path via the explicit `/gitlab set-default-instance` command.
- The `setDefaultInstance` logic itself is unchanged; only the authorization wrapper was added.
- Test coverage was appropriately updated to verify the new behavior.

**QA Recommendation:** 
Manual QA should focus on the complete setup and OAuth flow sequences with both admin and non-admin users:
1. Verify non-admin users cannot set default instance during setup (new error path)
2. Verify admin users can still set default instance during setup
3. Verify `/gitlab connect` returns OAuth link without any configuration changes
4. Verify the explicit `/gitlab set-default-instance` command still works for admins
5. Test bypass paths when users decline setting default instance in setup flow

Manual QA is recommended due to the authorization gate affecting user workflows, though the scope is bounded to instance setup/onboarding paths.

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -342,13 +342,6 @@ func (p *Plugin) handleConnect(args *model.CommandArgs, parameters []string) (*m
 		return p.getCommandResponse(args, "Please specify the instance name.", true), nil
 	}
 
-	// Set the default instance for the user before connecting
-	instanceName := strings.TrimSpace(strings.Join(parameters, " "))
-	err := p.setDefaultInstance(instanceName)
-	if err != nil {
-		return p.getCommandResponse(args, err.Error(), true), nil
-	}
-
 	pluginURL := getPluginURL(p.client)
 	if pluginURL == "" {
 		return p.getCommandResponse(args, "Encountered an error connecting to GitLab.", true), nil
--- a/server/command_test.go
+++ b/server/command_test.go
@@ -764,17 +764,10 @@ func TestInstanceCommands(t *testing.T) {
 	t.Run("connect", func(t *testing.T) {
 		for _, tc := range instanceNameTestCases {
 			t.Run("connect "+tc.name, func(t *testing.T) {
-				instanceList := []string{tc.instanceName}
-				if !tc.exists {
-					instanceList = []string{"Other Instance"}
-				}
-				p, msg, _ := setupInstanceCommandTest(t, instanceList, nil)
+				p, msg, api := setupInstanceCommandTest(t, nil, nil)
 				_, _ = p.handleConnect(args, tc.parameters)
-				if tc.exists {
-					assert.Contains(t, *msg, "Click here to link your GitLab account")
-				} else {
-					assert.Contains(t, *msg, "does not exist")
-				}
+				assert.Contains(t, *msg, "Click here to link your GitLab account")
+				api.AssertNotCalled(t, "SavePluginConfig", mock.Anything)
 			})
 		}
 		t.Run("no parameters", func(t *testing.T) {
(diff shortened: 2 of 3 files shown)


server/command.go
```diff
@@ -342,13 +342,6 @@ func (p *Plugin) handleConnect(args *model.CommandArgs, parameters []string) (*m
 		return p.getCommandResponse(args, "Please specify the instance name.", true), nil
```
edgarbellot, line 342: The flow framework's HTTP handler ([handler.go:18-24](https://github.com/mattermost/mattermost-plugin-gitlab/blob/master/server/flow.go#L820-L838)) uses the `Mattermost-User-ID` header to identify the caller but doesn't re-verify that the user has System Admin privileges before executing the step's `OnClick` callback. This means [stepSetDefaultInstance](https://github.com/mattermost/mattermost-plugin-gitlab/blob/master/server/flow.go#L820-L838) (820-838) relies entirely on the assumption that only admins can initiate the setup/oauth flows, rather than explicitly checking permissions at the point where `setDefaultInstance` is called.

In practice, exploiting this would require a non-admin user to craft a POST request with valid flow state, which isn't trivial. Still, would it make sense to add an explicit `isAuthorizedSysAdmin` check inside the `OnClick` handler at [line 826](https://github.com/mattermost/mattermost-plugin-gitlab/blob/master/server/flow.go#L826) before calling [setDefaultInstance](https://github.com/mattermost/mattermost-plugin-gitlab/blob/master/server/instance.go#L170-L197) (170-197)? This would follow the defense-in-depth pattern already used in [handleInstance](https://github.com/mattermost/mattermost-plugin-gitlab/blob/master/server/command.go#L232-L241) (232-241).

----

Related to the above - [submitDelegateSelection](https://github.com/mattermost/mattermost-plugin-gitlab/blob/master/server/flow.go#L318-L341) (318-341) allows an admin to delegate the setup wizard to any user via [StartSetupWizard](https://github.com/mattermost/mattermost-plugin-gitlab/blob/master/server/flow.go#L333) (333). The delegated user then receives the full flow including [stepSetDefaultInstance](https://github.com/mattermost/mattermost-plugin-gitlab/blob/master/server/flow.go#L820-L838) (820-838), which modifies the global plugin config for all users - without the delegate necessarily being an admin themselves.

Is this intentional? If so, it might be worth adding a comment clarifying that delegated users are trusted to perform admin-level config changes. If not, should the delegate's permissions be checked before presenting the "set as default instance" step?

Just flagging it for clarity since `setDefaultInstance` writes to the global config via [SavePluginConfig](https://github.com/mattermost/mattermost-plugin-gitlab/blob/master/server/instance.go#L190) (190).

edgarbellot (approved):
@avasconcelos114 LGTM :) I left a couple of questions. Making it non-blocking since I am not 100% sure if it's needed.

server/command.go
avasconcelos114: Ah thank you for the detailed review 🙌  
I'll add the defense-in-depth pattern for both consistency and to make the scenario you mentioned not just unlikely but simply impossible

nevyangelova (approved):
LGTM

PR #664, opened 2026-04-02
MM-67613 Adding encryption key rotation with graceful fallback

avasconcelos114:
#### Summary

This PR applies a more graceful behavior for rotation of encryptionkey values in the settings.

This PR adds a rotation of encrypted values alongside a fallback if the rotation fails that clear user sessions and notifies the user via DM. It also fills a gap in backwards compatibility, where a user is unable to run `/gitlab connect` unless instances have been installed, even when the global plugin settings have all the correct values set up

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-67613

#### QA Steps
The two main flows for this are for either a successful or a failed rotation, with:
**Success case**
1. Install the build with rotation logic present
2. Connect an account to a Gitlab instance
3. Go on the System Console and regenerate the encryption key for the Gitlab plugin
4. Exit the system console and run a gitlab plugin command (e.g. `/gitlab me`) to confirm things continue to work normally

**Failed rotation case**
1. Install a build _without_ rotation logic
2. Connect an account to a gitlab instance
3. regenerate the encryption key for the gitlab plugin
4. exit the system console and run a plugin command to confirm that the command fails due to auth errors
5. Install a build _with_ the rotation logic
6. Regenerate the encryption key and confirm through the system logs that the rotation failed
7. Exit the system console and confirm that a DM has been received by the plugin's bot notifying that the user has to run `/gitlab connect` in order to reconnect accounts
8. Running `/gitlab connect` works normally and the user is able to connect their account and continue to use the plugin normally

## Change Impact: 🟠 Medium

**Reasoning:** The PR modifies encryption/decryption, token re-encryption, and session/disconnect behavior — areas tied to authentication and data migration — but includes extensive unit tests and a clear fallback that limits broad failures.

**Regression Risk:** Medium — changes touch KV storage, crypto (padding/unpad), session cleanup, and connection gating. Tests cover many flows (including success, multiple failure modes, and fallback), but real-world migration edge cases (partial KV/CAS failures, multi-node concurrency, and forced disconnects) could surface in production.

** QA Recommendation:** Targeted manual QA required: (1) verify normal operations with no rotation; (2) validate successful rotation end-to-end (tokens re-encrypted, users remain connected); (3) simulate rotation failure and confirm affected users are disconnected, receive the plugin DM, and can reconnect via /gitlab connect. Given session-impacting behavior, do not skip manual QA.

*Generated by CodeRabbitAI*

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -338,15 +338,13 @@ func (p *Plugin) handleAbout(args *model.CommandArgs, parameters []string) (*mod
 }
 
 func (p *Plugin) handleConnect(args *model.CommandArgs, parameters []string) (*model.CommandResponse, *model.AppError) {
-	if len(parameters) < 1 {
-		return p.getCommandResponse(args, "Please specify the instance name.", true), nil
-	}
-
-	// Set the default instance for the user before connecting
-	instanceName := strings.TrimSpace(strings.Join(parameters, " "))
-	err := p.setDefaultInstance(instanceName)
-	if err != nil {
-		return p.getCommandResponse(args, err.Error(), true), nil
+	if len(parameters) >= 1 {
+		instanceName := strings.TrimSpace(strings.Join(parameters, " "))
+		if err := p.setDefaultInstance(instanceName); err != nil {
+			return p.getCommandResponse(args, err.Error(), true), nil
+		}
+	} else if !p.canConnect() {
+		return p.getCommandResponse(args, "No instance is configured. Please specify an instance name or ask your system administrator to configure the plugin.", true), nil
 	}
 
 	pluginURL := getPluginURL(p.client)
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -40,6 +40,11 @@ type configuration struct {
 	EnableCodePreview                string `json:"enablecodepreview"`
 	UsePreregisteredApplication      bool   `json:"usepreregisteredapplication"`
 	EnableChildPipelineNotifications bool   `json:"enablechildpipelinenotifications"`
+
+	// PreviousEncryptionKey is set internally during key rotation so that token
+	// reads can fall back to the old key while background re-encryption runs.
+	// It is never persisted to the plugin settings.
+	PreviousEncryptionKey string `json:"-"`
 }
 
 // Clone shallow copies the configuration. Your implementation may require a deep copy if
@@ -208,12 +213,17 @@ func (p *Plugin) OnConfigurationChange() error {
 	p.configurationLock.RUnlock()
 	newGitlabGroup := strings.TrimSpace(configuration.GitlabGroup)
 
-	p.setConfiguration(configuration, serverConfiguration)
-
 	if previousEncryptionKey != "" && configuration.EncryptionKey != "" &&
 		previousEncryptionKey != configuration.EncryptionKey {
+		configuration.PreviousEncryptionKey = previousEncryptionKey
+	}
+
+	p.setConfiguration(configuration, serverConfiguration)
+
+	if configuration.PreviousEncryptionKey != "" {
 		newKey := configuration.EncryptionKey
-		go p.reEncryptUserData(newKey, previousEncryptionKey)
+		prevKey := configuration.PreviousEncryptionKey
+		go p.reEncryptUserData(newKey, prevKey)
 	}
 
 	if hadConfig && p.BotUserID != "" && newGitlabGroup != "" && newGitlabGroup != previousGitlabGroup {
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -200,13 +200,21 @@ func (p *Plugin) OnConfigurationChange() error {
 	p.configurationLock.RLock()
 	hadConfig := p.configuration != nil
 	var previousGitlabGroup string
+	var previousEncryptionKey string
 	if hadConfig {
 		previousGitlabGroup = strings.TrimSpace(p.configuration.GitlabGroup)
+		previousEncryptionKey = p.configuration.EncryptionKey
 	}
 	p.configurationLock.RUnlock()
 	newGitlabGroup := strings.TrimSpace(configuration.GitlabGroup)
 
 	p.setConfiguration(configuration, serverConfiguration)
+
+	if previousEncryptionKey != "" && configuration.EncryptionKey != "" &&
+		previousEncryptionKey != configuration.EncryptionKey {
+		newKey := configuration.EncryptionKey
+		go p.reEncryptUserData(newKey, previousEncryptionKey)
+	}
 
 	if hadConfig && p.BotUserID != "" && newGitlabGroup != "" && newGitlabGroup != previousGitlabGroup {
 		p.notifyUsersOfDisallowedSubscriptions()
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -276,6 +276,22 @@ func (p *Plugin) getOAuthConfig() (*oauth2.Config, error) {
 	}, nil
 }
 
+// canConnect checks whether the plugin has enough configuration to allow a user to connect,
+// either via a KV-backed instance (default instance) or via legacy plugin settings.
+func (p *Plugin) canConnect() bool {
+	config := p.getConfiguration()
+
+	if config.DefaultInstanceName != "" {
+		if _, err := p.getInstance(config.DefaultInstanceName); err == nil {
+			return true
+		}
+	}
+
+	return config.GitlabOAuthClientID != "" &&
+		config.GitlabOAuthClientSecret != "" &&
+		isValidURL(config.GitlabURL) == nil
+}
+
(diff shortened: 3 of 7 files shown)


server/command.go
```diff
@@ -338,15 +338,13 @@ func (p *Plugin) handleAbout(args *model.CommandArgs, parameters []string) (*mod
...
-	// Set the default instance for the user before connecting
-	instanceName := strings.TrimSpace(strings.Join(parameters, " "))
-	err := p.setDefaultInstance(instanceName)
-	if err != nil {
-		return p.getCommandResponse(args, err.Error(), true), nil
+	if len(parameters) >= 1 {
+		instanceName := strings.TrimSpace(strings.Join(parameters, " "))
+		if err := p.setDefaultInstance(instanceName); err != nil {
+			return p.getCommandResponse(args, err.Error(), true), nil
+		}
```
avasconcelos114: Note: This setting of default instance behavior is going to be removed once https://github.com/mattermost/mattermost-plugin-gitlab/pull/663 is merged and this PR syncs to the master branch

server/configuration.go
```diff
@@ -200,14 +200,22 @@ func (p *Plugin) OnConfigurationChange() error {
...
 	}
 	p.configurationLock.RUnlock()
 	newGitlabGroup := strings.TrimSpace(configuration.GitlabGroup)
 
 	p.setConfiguration(configuration, serverConfiguration)
 
+	if previousEncryptionKey != "" && configuration.EncryptionKey != "" &&
+		previousEncryptionKey != configuration.EncryptionKey {
+		newKey := configuration.EncryptionKey
+		go p.reEncryptUserData(newKey, previousEncryptionKey)
```
nevyangelova: reEncryptUserData runs in a goroutine. If OnConfigurationChange fires again before re-encryption completes a second goroutine could start with a different previousEncryptionKey that is actually the new key from the first rotation and data corruption. Consider adding a check inside reEncryptUserData to verify that previousEncryptionKey still matches what's in the KV store

server/plugin.go
```diff
@@ -991,6 +1011,158 @@ func (p *Plugin) handleRevokedToken(info *gitlab.UserInfo) {
...
+	var tokenBytes []byte
+	if err := p.client.KV.Get(kvKey, &tokenBytes); err != nil || tokenBytes == nil {
+		return false, nil
+	}
+
+	tokenStr := string(tokenBytes)
+
+	// Idempotency check: if the token can already be decrypted with the new key, it was already migrated.
+	if _, err := decrypt([]byte(newEncryptionKey), tokenStr); err == nil {
+		return false, nil
```
nevyangelova: There's a small chance decrypting with the wrong key produces valid padding, causing a false-positive. Consider adding a version prefix to re-encrypted values instead.

server/plugin.go
```diff
@@ -991,6 +1011,158 @@ func (p *Plugin) handleRevokedToken(info *gitlab.UserInfo) {
...
+
+	plainJSON, err := decrypt([]byte(previousEncryptionKey), tokenStr)
+	if err != nil {
+		p.client.Log.Warn("Failed to decrypt token with previous key during re-encryption, force-disconnecting user",
+			"user_id", userID, "error", err.Error())
+		p.forceDisconnectUser(userID)
+		return false, err
+	}
+
+	reEncrypted, err := encrypt([]byte(newEncryptionKey), plainJSON)
```
nevyangelova, line 1132: If decrypt returns garbage that passes padding corrupted data gets re encrypted silently. Add json.Unmarshal validation before re-encrypting.

server/plugin.go
```diff
@@ -991,6 +1011,158 @@ func (p *Plugin) handleRevokedToken(info *gitlab.UserInfo) {
...
+			result.ForceDisconnectCount++
+			continue
+		}
+		if migrated {
+			result.MigratedCount++
+		}
+	}
+
+	auditRec.Success()
+	auditRec.AddEventResultState(result)
```
nevyangelova, line 1073: Even if ForceDisconnectCount == len(allKeys) audit says success. Mark as failure orpartial when disconnects occurred.

nevyangelova (changes requested):
Thanks @avasconcelos114 a few comments!

server/configuration.go
```diff
@@ -200,14 +200,22 @@ func (p *Plugin) OnConfigurationChange() error {
...
 	var previousGitlabGroup string
+	var previousEncryptionKey string
 	if hadConfig {
 		previousGitlabGroup = strings.TrimSpace(p.configuration.GitlabGroup)
+		previousEncryptionKey = p.configuration.EncryptionKey
 	}
 	p.configurationLock.RUnlock()
 	newGitlabGroup := strings.TrimSpace(configuration.GitlabGroup)
 
 	p.setConfiguration(configuration, serverConfiguration)
```
edgarbellot: There's a window between [setConfiguration](https://github.com/mattermost/mattermost-plugin-gitlab/blob/7fab3d923e820325dafce13640037666e8670813/server/configuration.go#L211) (211) activating the new key and [reEncryptUserData](https://github.com/mattermost/mattermost-plugin-gitlab/blob/7fab3d923e820325dafce13640037666e8670813/server/configuration.go#L216) (216) finishing. During that window, [getGitlabUserTokenByMattermostID](https://github.com/mattermost/mattermost-plugin-gitlab/blob/7fab3d923e820325dafce13640037666e8670813/server/plugin.go#L476-L497) (476-497) tries to decrypt with the new key, but the tokens are still encrypted with the old one - so every user gets a decryption failure until their individual token is re-encrypted. In deployments with many users this could be a noticeable outage.

Would it make sense to add a fallback decryption attempt with the previous key inside [getGitlabUserTokenByMattermostID](https://github.com/mattermost/mattermost-plugin-gitlab/blob/7fab3d923e820325dafce13640037666e8670813/server/plugin.go#L486-L489) (486-489)? Something like storing the previous key temporarily in the config (e.g. PreviousEncryptionKey) and, if decrypt with the current key fails, retrying with the previous one before returning an error. That way users stay functional while the background migration runs.

An alternative would be to run the re-encryption synchronously before activating the new key, but that could block `OnConfigurationChange` for a while depending on user count, so the fallback approach seems more practical.

edgarbellot (changes requested):
Thank you @avasconcelos114 ! I just have one comment :)

server/plugin.go
avasconcelos114: I think this also resolves the false-positive issue you listed above so I'll keep this as a way to resolve both, thanks for the suggestion!

server/plugin.go
avasconcelos114: Oh yes this makes more sense, I'll adjust this

server/configuration.go
avasconcelos114: Ah to prevent such a scenario I applied `cluster.NewMutex` to ensure only one goroutine handles the re-encryption and we don't end up with two competing goroutines, and in the event that we have two quick succession of migrations happening, the `forceDisconnectUser` flow should kick in and notify users to reconnect their accounts

server/configuration.go
avasconcelos114: Ah yep that makes sense to me, I'll add this temporary config value as a fallback and this should make the process smoother, thanks for the suggestion!

server/configuration.go
```diff
@@ -40,6 +40,11 @@ type configuration struct {
 	EnableCodePreview                string `json:"enablecodepreview"`
 	UsePreregisteredApplication      bool   `json:"usepreregisteredapplication"`
 	EnableChildPipelineNotifications bool   `json:"enablechildpipelinenotifications"`
+
+	// PreviousEncryptionKey is set internally during key rotation so that token
+	// reads can fall back to the old key while background re-encryption runs.
+	// It is never persisted to the plugin settings.
+	PreviousEncryptionKey string `json:"-"`
```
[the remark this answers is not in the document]
avasconcelos114: Considering that the likelihood of an encryption key rotation (which is something that normally doesn't happen) to also be forcefully interrupted halfway is exceedingly low. As well as the fact that the force disconnect fallback in the `/gitlab disconnect` command has also been introduced, makes me feel like this added complexity is not worth it for managing an edge case of something that is already an edge-case

I think we can skip this one

avasconcelos114:
@coderabbitai approve

edgarbellot (approved):
Perfectly addressed, as always! Thank you Andre!

nevyangelova (commented):
LGTM!

nevyangelova (approved):
LGTM!

ogi-m (approved):
LGTM!

avasconcelos114:
@nang2049 Could you approve this PR again? GitHub disabled the merge button with a message that it needs an approval from `mattermost/integrations` code owners

nang2049 (approved):

PR #678, opened 2026-06-11
MM-69161 Adding permission check model to allow/deny requests

avasconcelos114:
#### Summary

This PR adds a layer that checks gitlab's own permission model to prevent any state-changing requests from being made, and failing that it falls back to the current flow for only allowing sysadmins from accessing a particular command. This has been for now only integrated with the `/gitlab webhook` commands

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-69161

## Change Impact: 🟠 Medium

**Reasoning:** The PR changes authorization behavior for `/gitlab webhook` by removing a local sysadmin gate and relying on GitLab permission enforcement plus namespace allow/deny checks, and it extends audit logging for webhook creation. While this is security-relevant, the blast radius is primarily confined to webhook command flows and is supported by newly added namespace-denial tests.

**Regression Risk:** Medium—command access semantics change (admin-only → namespace-scoped/remote-permission-scoped), and new early-denial paths can alter behavior for existing configurations; however, the affected logic is localized and key denial/success scenarios are covered by tests.

**QA Recommendation:** Moderate manual QA. Validate for both admin and non-admin users:
1) `/gitlab webhook list` for allowed vs disallowed namespaces (including denial messaging and no hook retrieval),
2) `/gitlab webhook add` for allowed vs disallowed namespaces (including denial messaging),
3) audit logging on successful webhook add (hook id/url/scope present) and on forbidden failures (failure recorded correctly).
*Generated by CodeRabbitAI*


avasconcelos114:
@edgarbellot This PR here adds the general flow we had previously discussed, but now that I see it in action I can't help but wonder if we should just remove the admin check from the webhook commands instead as Gitlab's own API would be rejecting a call made with the oauth token of a user that doesn't have the needed permissions anyway :D 

Unless we see any value in preventing an API call from being made to Gitlab until we're sure the user has due access beforehand, but for now I kept the changes and flow until we reach a decision, it would be quite quick and easy to revert the previous commit and add a new one that doesn't enforce sys admin checks for these two commands

server/gitlab/permissions.go
```diff
@@ -0,0 +1,69 @@
...
+		fmt.Sprintf("%s/%s", namespace, project),
+		user.GitlabUserID,
+		internGitlab.WithContext(ctx),
+	)
+	if respErr := checkResponse(resp); respErr != nil {
+		return internGitlab.NoPermissions, respErr
+	}
+	if err != nil {
+		return internGitlab.NoPermissions, err
+	}
```
nang2049: nit: If the request errored out at the transport layer resp could be nil here depending on checkResponse's contract. I would swap the order so you check err != nil first when resp == nil. Same pattern repeats in GetGroupAccessLevel.

server/gitlab/permissions.go
```diff
@@ -0,0 +1,69 @@
...
+// The go-gitlab client does not expose a single-user inherited group lookup, so
+// this issues a raw request to GET /groups/:id/members/all/:user_id. ErrNotFound
+// is returned when the user has no membership at all.
+func (g *gitlab) GetGroupAccessLevel(ctx context.Context, user *UserInfo, token *oauth2.Token, group string) (internGitlab.AccessLevelValue, error) {
+	client, err := g.GitlabConnect(*token)
+	if err != nil {
+		return internGitlab.NoPermissions, err
+	}
+
+	endpoint := fmt.Sprintf("groups/%s/members/all/%d", internGitlab.PathEscape(group), user.GitlabUserID)
```
nang2049: Nit: here we use internGitlab.PathEscape for the group identifier but GetProjectAccessLevel above does fmt.Sprintf("%s/%s", namespace, project) with no escaping. Can we align both to use PathEscape for symmetry?

server/permissions.go
```diff
@@ -0,0 +1,99 @@
...
+		if errors.Is(err, gitlab.ErrNotFound) {
+			return false, nil
+		}
+		p.client.Log.Warn("Falling back to System Admin check after GitLab permission lookup failed", "err", err.Error())
+		return p.isAuthorizedSysAdmin(info.UserID)
+	}
+
+	required, ok := minimumAccessLevels[action][scope]
+	if !ok {
+		return false, errors.Errorf("no minimum access level configured for action %q scope %q", action, scope)
```
nang2049: Consider making this a build time guarantee like a func init() that verifies every permissionAction constant has a complete Project/Group entry in minimumAccessLevels. That way its caught at plugin boot isntead  when we run the command in production.

server/permissions.go
```diff
@@ -0,0 +1,99 @@
...
+// permissions to manage webhooks. Returns an empty string on success or the
+// user-facing denial message.
+func (p *Plugin) requireWebhookPermission(ctx context.Context, info *gitlab.UserInfo, group, project string) string {
+	allowed, err := p.hasGitlabPermission(ctx, info, group, project, actionManageWebhook)
+	if err != nil {
+		p.client.Log.Warn("Failed to check GitLab permission for webhook", "err", err.Error())
+	}
+	if !allowed {
+		return insufficientPermissionsMessage
+	}
```
nang2049: The err is logged but then discarded when allowed's zero value is false so a misconfig silently surfaces to the user as insufficientPermissionsMessage which might be confusing UX

server/command.go
```diff
@@ -1137,7 +1136,6 @@ func (p *Plugin) getAutocompleteData(config *configuration) *model.AutocompleteD
 	gitlab.AddCommand(settings)
 
 	webhook := model.NewAutocompleteData("webhook", "[command]", "Available Commands: list, add")
-	webhook.RoleID = model.SystemAdminRoleId
```
nang2049, line 1140: Would this mean that if a non-Maintainer will see and try the command they will hit the denial message?

server/permissions_test.go
```diff
@@ -0,0 +1,110 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
```
nang2049: Nice coverage on the matrix. I would add a couple cases for full safety but living up to your consideration! 

1. Action with no min-access-level configured which exercises the errors.Errorf("no minimum access level configured for action %q scope %q", ...) branch.
2. requireWebhookPermission directly as currently only hasGitlabPermission is unit-tested.

nang2049 (commented):
Amazing change thanks @avasconcelos114 a few minor comments!

server/command.go
```diff
@@ -662,6 +657,10 @@ func (p *Plugin) webhookCommand(ctx context.Context, parameters []string, info *
 			return namespaceErr.Error()
 		}
 
+		if msg := p.requireWebhookPermission(ctx, info, group, project); msg != "" {
```
edgarbellot: nit: webhook add creates a state-changing resource but doesn't emit an audit record, unlike other state-changing ops in the repo ([createIssue](https://github.com/mattermost/mattermost-plugin-gitlab/blob/24a987e08a97ccdfbe1db988a08a334cbce9a3ad/server/api.go#L641-L650)). Since you're touching the auth flow here it could be a good spot to add a `MakeAuditRecord` / `LogAuditRec` capturing the actor and target. Non-blocking :)

edgarbellot:
> @edgarbellot This PR here adds the general flow we had previously discussed, but now that I see it in action I can't help but wonder if we should just remove the admin check from the webhook commands instead as Gitlab's own API would be rejecting a call made with the oauth token of a user that doesn't have the needed permissions anyway :D
> 
> Unless we see any value in preventing an API call from being made to Gitlab until we're sure the user has due access beforehand, but for now I kept the changes and flow until we reach a decision, it would be quite quick and easy to revert the previous commit and add a new one that doesn't enforce sys admin checks for these two commands

I agree, let's remove the check and rely on GitLab.

Since these calls use the user's own OAuth token, GitLab is already the source of truth and rejects anything they can't do. That holds for sysadmins too.

edgarbellot (approved):
Looking good! just one nit. Thanks Andre!

server/command.go
avasconcelos114: Ah yep sounds good! so for this PR I'll now:
1. Revert the previous PR with the explicit permission checks
2. Remove the sysadmin checks for webhook commands in order to allow Gitlab's own permission checks to kick in (with some graceful error handling so it's clear why was the call rejected)
3. Add audit logging for the webhook command handlers

edgarbellot: > with some graceful error handling so it's clear why was the call rejected

good call!

Sounds good - happy to review the changes once they are ready.

avasconcelos114:
Made the changes necessary to simplify this quite a bit by simply allowing Gitlab to reject the request made with the user's oauth token and removed the sysadmin checks previously in place.

I've also gone ahead and enforced the same checks we had in other commands for namespace limiting via the `GitLab Group:` option + Added audit logging for the transaction

server/command.go
```diff
@@ -662,10 +671,40 @@ func (p *Plugin) webhookCommand(ctx context.Context, parameters []string, info *
...
+
+		auditParams := AddWebhookAuditParams{
+			MattermostUserID: info.UserID,
+			GitlabUsername:   info.GitlabUsername,
+			Namespace:        resolvedNamespace,
+			URL:              hookOptions.URL,
+		}
+		model.AddEventParameterAuditableToAuditRec(auditRec, "add_webhook", auditParams)
+
+		if err := p.isNamespaceAllowed(resolvedNamespace); err != nil {
```
edgarbellot, line 688: Thanks for this!

edgarbellot (approved):
Looking good :)

nang2049 (approved):
LGTM

PR #689, opened 2026-07-30
MM-69953 Fixing inconsistent behavior for issue webhook events

avasconcelos114:
#### Summary

This PR closes a gap in behaviors that happen across different types of issues and how webhooks handle them

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-69953

## Change Impact: 🟠 Medium

**Reasoning:** The change updates a shared webhook subscription API across multiple handlers and adds authorization checks for confidential issues. Tests cover confidential and non-confidential notification paths, which reduces regression risk.

**Regression Risk:** Medium—incorrect confidentiality propagation or permission handling could affect notification delivery across several webhook event types.

**QA Recommendation:** Perform targeted manual QA for confidential issues, issue comments, and group/project subscriptions. Verify that non-confidential webhook events remain unchanged. Skipping manual QA carries moderate risk.

*Generated by CodeRabbitAI*

### DIFF
--- a/server/webhook.go
+++ b/server/webhook.go
@@ -57,8 +57,9 @@ func (g *gitlabRetreiver) GetSubscribedChannelsForProject(
 	namespace string,
 	project string,
 	isPublicVisibility bool,
+	isConfidential bool,
 ) []*subscription.Subscription {
-	return g.p.GetSubscribedChannelsForProject(ctx, namespace, project, isPublicVisibility)
+	return g.p.GetSubscribedChannelsForProject(ctx, namespace, project, isPublicVisibility, isConfidential)
 }
 
 func (p *Plugin) handleWebhook(w http.ResponseWriter, r *http.Request) {
--- a/server/webhook.go
+++ b/server/webhook.go
@@ -220,17 +220,47 @@ func (p *Plugin) sendRefreshIfNotAlreadySent(alreadySentRefresh map[string]bool,
 }
 
 func (p *Plugin) permissionToProject(ctx context.Context, userID, namespace, project string) bool {
-	if userID == "" {
+	result := p.fetchProjectAsUser(ctx, userID, namespace, project)
+	if result == nil {
 		return false
 	}
 
-	if err := p.isNamespaceAllowed(namespace); err != nil {
+	return effectiveAccess(result.Permissions) > gitlabLib.GuestPermissions
+}
+
+// permissionToSubscribe reports whether the user may create a subscription for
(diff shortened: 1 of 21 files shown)


avasconcelos114:
@coderabbit approve if the last commit addressed the previously raised concerns

server/webhook.go
jgheithcock, line 297: The core issue seems to still be here, if both ProjectAccess and GroupAccess is nil, you skip the checks. The recommendation is to add a helper to handle this:
```
func effectiveAccess(perms *gitlabLib.Permissions) gitlabLib.AccessLevelValue {
	if perms == nil {
		return 0
	}
	var level gitlabLib.AccessLevelValue
	if perms.ProjectAccess != nil && perms.ProjectAccess.AccessLevel > level {
		level = perms.ProjectAccess.AccessLevel
	}
	if perms.GroupAccess != nil && perms.GroupAccess.AccessLevel > level {
		level = perms.GroupAccess.AccessLevel
	}
	return level
}

// For confidential content, require > Guest
// return effectiveAccess(result.Permissions) > gitlabLib.GuestPermissions
```

I definitely think this is more readable and more correct

jgheithcock (changes requested):

avasconcelos114:
Note: Testing the cherry-pick automation - I'll be closing any PRs prematurely created as part of the tests

server/webhook/note.go
```diff
@@ -67,13 +67,18 @@ func (w *webhook) handleChannelIssueComment(ctx context.Context, event *gitlab.I
 	subs := w.gitlabRetreiver.GetSubscribedChannelsForProject(
 		ctx, namespace, project,
 		repo.Visibility == gitlab.PublicVisibility,
+		event.Issue.Confidential,
```
jgheithcock: I think CodeRabbit didn't quite go far enough here - sorry I missed it. We should use the same metric for the comments as for handleChannelIssue and add

`isConfidential := event.Issue.Confidential || event.EventType == string(gitlab.EventConfidentialNote)`
above and use `isConfidential` in place of the bare `event.Issue.Confidential` (So we catch it either way.)

We would also want to add a regression test where event_type is confidential_note and issue.confidential is absent/false.

jgheithcock (approved):
I'm marking this approved so you don't need a re-review, but I do think we should be consistent here (as well as catching this edge case).

hmohammed-prodsec (approved):
Thanks for the changes, LGTM 👍

PR #691, opened 2026-08-02
Fix lost-update race in channel subscriptions

aegonmyy:
# Fix lost-update race in channel subscriptions

## Summary

All channel subscriptions are stored in a single KV key (`subscriptions`).
Every mutation reads the whole blob, changes it in memory, and writes the whole
blob back. That read-modify-write was not atomic, so two subscribe or
unsubscribe operations running at the same time in different channels could
overwrite each other. Both callers were told the operation succeeded, but only
one of the two changes actually persisted. The other was silently lost.

## Root cause

`StoreSubscriptions` in `server/subscriptions.go` wrote the whole blob with a
plain `Set` and no compare-and-set, and it also swallowed any write error:

```go
// server/subscriptions.go (before)
func (p *Plugin) StoreSubscriptions(s *Subscriptions) error {
	if _, err := p.client.KV.Set(SubscriptionsKey, s); err != nil {
		p.client.Log.Warn("can't set subscriptions in kvstore", "err", err.Error())
	}
	return nil   // always reports success, even when the write failed
}
```

`AddSubscription` (`subscriptions.go:65`) and `Unsubscribe`
(`subscriptions.go:159`) both did `GetSubscriptions()` to read the whole blob,
mutated the in-memory copy, then handed that copy to `StoreSubscriptions`.

Because the write used a plain `Set`, a concurrent writer that had read the same
starting blob would overwrite the first writer's change with its own copy. The
last write won and every earlier change was dropped. Since the error was
swallowed, a genuinely failed write also reported success.

## The fix

Switch to the atomic `SetAtomicWithRetries` pattern exposed by the pluginapi KV
service, and move the read-modify-write inside its callback. A new
`modifySubscriptions` helper decodes `oldValue` on every attempt, applies a
mutate closure to that fresh state, and returns the re-encoded result. On a
conflicting write the retry re-reads the latest blob and re-applies the change,
so nothing is lost. The real store error is now returned to the caller instead
of being swallowed.

`AddSubscription` and `Unsubscribe` now express their change as a mutate closure
and keep their existing return signatures (`Unsubscribe` still reports whether a
subscription was removed and returns the resulting subscriptions).
`GetSubscriptions` and `GetSubscriptionsByChannel` are unchanged.

## Regression test

`TestSubscriptionRace` (`server/subscription_race_test.go`) fires 200 concurrent
`AddSubscription` calls against an in-memory KV store whose `KVGet` and
`KVSetWithOptions` implement real compare-and-set semantics, matching the server
KV store. It then compares how many calls reported success against how many
subscriptions actually persisted. It fails on the old code and passes with this
fix.

### Before the fix

```
=== RUN   TestSubscriptionRace
    subscription_race_test.go:106: concurrent AddSubscription calls : 200
    subscription_race_test.go:107: AddSubscription returned success : 200
    subscription_race_test.go:108: subscriptions actually persisted : 10
    subscription_race_test.go:109: silently lost (success but gone) : 190
    subscription_race_test.go:112: lost-update bug: 190 subscriptions were reported as saved but silently dropped
--- FAIL: TestSubscriptionRace (0.00s)
FAIL
```

All 200 calls reported success because the old code swallowed write errors, yet
only 10 subscriptions persisted.

### After the fix

```
=== RUN   TestSubscriptionRace
    subscription_race_test.go:106: concurrent AddSubscription calls : 200
    subscription_race_test.go:107: AddSubscription returned success : 123
    subscription_race_test.go:108: subscriptions actually persisted : 123
    subscription_race_test.go:109: silently lost (success but gone) : 0
--- PASS: TestSubscriptionRace (0.06s)
```

After the fix, every call that reports success is durable. Under heavy
contention some calls exhaust the five internal retries and return an error
instead. Those are reported as failures and are not counted as success, which is
the correct behavior.

## Change Impact: 🟠 Medium

**Reasoning:** The change modifies subscription persistence and concurrency handling. The scope is isolated, but it affects data integrity and removes the public `StoreSubscriptions` method.

**Regression Risk:** Medium. The atomic write flow changes `AddSubscription` and `Unsubscribe`, including error handling and removal behavior. A concurrency regression test covers concurrent additions, but other mutation paths remain less covered.

** QA Recommendation:** Perform focused manual QA for concurrent updates, replacements, unsubscriptions, store errors, and consumers of `StoreSubscriptions`. Skipping manual QA carries moderate risk because the change affects persisted subscription data.

*Generated by CodeRabbitAI*

### DIFF
--- a/server/subscription_race_test.go
+++ b/server/subscription_race_test.go
@@ -111,4 +111,11 @@ func TestSubscriptionRace(t *testing.T) {
 	if silentlyLost > 0 {
 		t.Fatalf("lost-update bug: %d subscriptions were reported as saved but silently dropped", silentlyLost)
 	}
+
+	// Guard against a false pass: if every AddSubscription call failed,
+	// reportedSuccess and persisted are both 0, so silentlyLost is 0 and the
+	// check above passes without actually exercising the race.
+	if reportedSuccess == 0 {
+		t.Fatalf("no AddSubscription calls succeeded; the race was not exercised")
+	}
 }
--- a/server/subscription_race_test.go
+++ b/server/subscription_race_test.go
@@ -1,0 +1,114 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"sync"
+	"testing"
+
+	"github.com/mattermost/mattermost/server/public/model"
+	"github.com/mattermost/mattermost/server/public/plugin/plugintest"
+	"github.com/mattermost/mattermost/server/public/pluginapi"
+
+	"github.com/mattermost/mattermost-plugin-gitlab/server/subscription"
+)
+
+// raceKVAPI is an in-memory plugin.API whose KVGet and KVSetWithOptions
+// implement real compare-and-set semantics, so SetAtomicWithRetries behaves
+// exactly as it does against the server KV store. Every other API method is
+// inherited from plugintest.API and is unused by the subscription store.
+type raceKVAPI struct {
+	plugintest.API
+	mu    sync.Mutex
+	store map[string][]byte
+}
+
+func newRaceKVAPI() *raceKVAPI {
+	return &raceKVAPI{store: map[string][]byte{}}
+}
+
+func (a *raceKVAPI) KVGet(key string) ([]byte, *model.AppError) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	return a.store[key], nil
+}
+
+func (a *raceKVAPI) KVSetWithOptions(key string, value []byte, options model.PluginKVSetOptions) (bool, *model.AppError) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	if options.Atomic {
+		if !bytes.Equal(a.store[key], options.OldValue) {
+			return false, nil
+		}
+	}
+
+	if value == nil {
+		delete(a.store, key)
+	} else {
+		a.store[key] = value
+	}
+	return true, nil
+}
+
+// TestSubscriptionRace drives the plugin's real AddSubscription against an
+// in-memory KV store with real compare-and-set semantics. Many channels
+// subscribe to the same repository concurrently; afterwards we compare how many
+// AddSubscription calls reported success (nil error) against how many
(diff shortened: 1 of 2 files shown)


server/subscriptions.go
```diff
@@ -161,41 +192,52 @@ func (p *Plugin) Unsubscribe(channelID string, fullPath string) (bool, *Subscrip
...
 	if !removed {
-		return false, subs, nil
+		return false, current, nil
 	}
-	return true, subs, p.StoreSubscriptions(subs)
+	if err != nil {
+		return false, nil, err
+	}
+
+	return true, subs, nil
```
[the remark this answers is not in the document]
aegonmyy: Good catch. Fixed: `Unsubscribe` now checks the error from `modifySubscriptions` before the removed flag. A read/unmarshal failure (which prevents the mutate callback from running) is surfaced as a real error, and only `errStopModify` is treated as the no-op "no matching subscription" case. Tests still pass.

server/subscription_race_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	silentlyLost := reportedSuccess - persisted
+	t.Logf("concurrent AddSubscription calls : %d", numChannels)
+	t.Logf("AddSubscription returned success : %d", reportedSuccess)
+	t.Logf("subscriptions actually persisted : %d", persisted)
+	t.Logf("silently lost (success but gone) : %d", silentlyLost)
+
+	if silentlyLost > 0 {
+		t.Fatalf("lost-update bug: %d subscriptions were reported as saved but silently dropped", silentlyLost)
+	}
```
jgheithcock, line 113: If all AddSubscription calls fail, `subs` will be empty and persisted will be 0 and so silentlyLost will be 0 and no failures will be reported. Suggest adding the following after silentlyLost check:
```
if reportedSuccess == 0 {
    t.Fatalf("No AddSubscription calls succeeded")
}
```

jgheithcock (changes requested):

server/subscription_race_test.go
aegonmyy: Thanks, addressed. Added the same guard after the silentlyLost check: the test now fails with "no AddSubscription calls succeeded" if reportedSuccess == 0, so it can't pass without actually exercising the race. Pushed in 8ab8a0e.

PR #692, opened 2026-08-03
Use constant-time comparison for GitLab webhook secret

aegonmyy:
## Summary
`handleWebhook` compares the configured `WebhookSecret` against the incoming `X-Gitlab-Token` header with a plain `!=` string comparison. Go's string inequality is not constant-time (it can return as soon as it hits a differing byte), which leaks timing information about the secret on an endpoint that has no other authentication.

Fix: use `subtle.ConstantTimeCompare`, matching the pattern the sibling GitHub plugin already uses (`hmac.Equal`) for its own webhook signature check.

## Test plan
- [x] Existing `TestHandleWebhookBadSecret` and the valid-secret webhook tests in `server/webhook_test.go` cover both branches; behavior is unchanged (only the comparison mechanism changed, not its outcome for equal/unequal inputs).

## Change Impact: 🟡 Medium

**Reasoning:** The change affects webhook authentication but remains isolated to one module. Existing tests cover valid and invalid secrets, and the unauthorized response remains unchanged.

**Regression Risk:** Low. The comparison uses fixed-length SHA-256 digests and preserves the existing validation result.

** QA Recommendation:** Focused manual QA is recommended for valid, invalid, and near-match `X-Gitlab-Token` values. Skipping manual QA presents low functional risk.

*Generated by CodeRabbitAI*

### DIFF
--- a/server/webhook.go
+++ b/server/webhook.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"context"
+	"crypto/subtle"
 	"fmt"
 	"io"
 	"net/http"
@@ -65,7 +66,7 @@ func (p *Plugin) handleWebhook(w http.ResponseWriter, r *http.Request) {
 	config := p.getConfiguration()
 
 	signature := r.Header.Get("X-Gitlab-Token")
-	if config.WebhookSecret != signature {
+	if subtle.ConstantTimeCompare([]byte(config.WebhookSecret), []byte(signature)) != 1 {
 		http.Error(w, "Not authorized", http.StatusUnauthorized)
 		return
 	}
--- a/server/webhook.go
+++ b/server/webhook.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"context"
+	"crypto/sha256"
 	"crypto/subtle"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


server/webhook.go
```diff
@@ -65,7 +66,7 @@ func (p *Plugin) handleWebhook(w http.ResponseWriter, r *http.Request) {
 	config := p.getConfiguration()
 
 	signature := r.Header.Get("X-Gitlab-Token")
-	if config.WebhookSecret != signature {
+	if subtle.ConstantTimeCompare([]byte(config.WebhookSecret), []byte(signature)) != 1 {
```
jgheithcock: The remaining issue is that, unlike `hmac.Equal()`, the request's 'X-Gitlab-Token' can vary in it length, and `subtle.ConstantTimeCompare` returns 0 immediately if the two strings are different lengths. (And so you can at least determine the length of the secret.) So it seems we should keep going with this and compare digests of these two, i.e.:
```
secret := sha256.Sum256([]byte(config.WebhookSecret))
token := sha256.Sum256([]byte(signature))
if subtle.ConstantTimeCompare(secret[:], token[:]) != 1 {
	http.Error(w, "Not authorized", http.StatusUnauthorized)
	return
}
```

jgheithcock (changes requested):

server/webhook.go
aegonmyy: You're right, ConstantTimeCompare returns 0 immediately on a length mismatch, which leaks the length of the configured secret. Switched to comparing fixed-length SHA-256 digests of the secret and the incoming token so the comparison is constant time regardless of input length. Pushed in 74dba17.

jgheithcock (approved):
LGTM, thanks!

PR #702, opened 2026-08-14
MM-67925 Added deduplication of DM notifications

avasconcelos114:
#### Summary

This PR adds a dedup mechanism similar to Jira's to the Gitlab plugin. If a DM with the same contents is sent to the same user in a short span of time, the plugin will prevent the second DM from being sent

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-67925

#### QA Notes
The easiest way to reproduce this is to create 2 webhook entries that target the same endpoint, this way the same action will cause two requests to be sent to the plugin

## Change Impact: 🟡 Medium

**Reasoning:** The change affects shared webhook deduplication for direct messages and channel posts, plus plugin DM channel error handling. Extensive tests cover concurrent delivery, KV failures, claim expiry, stale-token protection, and delivery failures.

**Regression Risk:** Incorrect claim handling could suppress valid notifications. KV failures fail open, and compare-and-delete protects newer claims from stale releases.

** QA Recommendation:** Perform targeted manual QA for duplicate and concurrent webhooks, KV failures, DM channel failures, channel-post failures, and normal delivery. Skipping manual QA has a moderate risk of missing integration issues.

*Generated by CodeRabbitAI*

### DIFF
--- a/server/webhook.go
+++ b/server/webhook.go
@@ -234,36 +234,45 @@ func channelPostDedupKey(channelID, message string) string {
 }
 
 // claimDedupKey atomically claims dedupKey: SetAtomic(nil) only writes when the
-// key is absent, so concurrent deliveries can't both win. A KV error fails open,
-// returning deliver without owned since nothing was written.
-func (p *Plugin) claimDedupKey(caller, dedupKey string) (deliver, owned bool) {
-	claimed, err := p.client.KV.Set(dedupKey, true,
+// key is absent, so concurrent deliveries can't both win. It returns the token
+// identifying this claim, which releaseDedupKey needs to prove ownership. A KV
+// error fails open, returning deliver with an empty token since nothing was
+// written.
+func (p *Plugin) claimDedupKey(caller, dedupKey string) (deliver bool, claimToken string) {
+	token := model.NewId()
+	claimed, err := p.client.KV.Set(dedupKey, []byte(token),
 		pluginapi.SetExpiry(webhookDedupTTL),
 		pluginapi.SetAtomic(nil),
 	)
 	switch {
 	case err != nil:
 		p.client.Log.Warn(caller+": failed to claim dedup key, delivering anyway", "dedup_key", dedupKey, "err", err.Error())
-		return true, false
+		return true, ""
 	case !claimed:
 		p.client.Log.Debug(caller+": another delivery already claimed this notification, skipping", "dedup_key", dedupKey)
-		return false, false
+		return false, ""
 	}
-	return true, true
+	return true, token
 }
 
-// releaseDedupKey lets a redelivery retry instead of waiting out the TTL. Only
-// call it when this delivery owns the claim and nothing was posted.
-func (p *Plugin) releaseDedupKey(caller, dedupKey string) {
-	if err := p.client.KV.Delete(dedupKey); err != nil {
+// releaseDedupKey lets a redelivery retry instead of waiting out the TTL. It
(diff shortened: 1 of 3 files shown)


server/webhook_test.go
```diff
@@ -191,3 +202,215 @@ func TestHandleWebhookForChildPipelineNotficationEnabled(t *testing.T) {
...
+	p.client = pluginapi.NewClient(api, p.Driver)
+
+	return p, api
+}
+
+func TestHandleWebhookDeduplicatesDuplicateDelivery(t *testing.T) {
+	p, api := setupDedupTestPlugin(t)
+
+	var claims atomic.Int32
+	api.On("KVSetWithOptions", mock.AnythingOfType("string"), mock.Anything, mock.Anything).Return(
```
nang2049: The options argument is stubbed as `mock.Anything` in all tests so nothing in the suite asserts the two properties the mechanissm depends on. I verified this by deleting them: 

- removing `pluginapi.SetAtomic(nil)` leaves `go test ./server/` green
- removing `pluginapi.SetExpiry(notificationDedupTTL)` also leaves the suite green.

A matcher can work here: 

```go
func isDedupClaimOptions(opts model.PluginKVSetOptions) bool {
    return opts.Atomic && opts.OldValue == nil &&
        opts.ExpireInSeconds == int64(notificationDedupTTL/time.Second)
}
```
Separately this stub also ignores the key and decides by call count so this test does not verify keying

server/webhook_test.go
```diff
@@ -191,3 +202,215 @@ func TestHandleWebhookForChildPipelineNotficationEnabled(t *testing.T) {
...
+	}
+
+	api.AssertNumberOfCalls(t, "CreatePost", 1)
+}
+
+func TestHandleWebhookDeduplicatesConcurrentDeliveries(t *testing.T) {
+	p, api := setupDedupTestPlugin(t)
+
+	var claims atomic.Int32
+	api.On("KVSetWithOptions", mock.AnythingOfType("string"), mock.Anything, mock.Anything).Return(
```
nang2049: Same as the stub in `TestHandleWebhookDeduplicatesDuplicateDelivery` the third argument should be an options matcher rather than `mock.Anything` otherwise this test passes even with `pluginapi.SetAtomic(nil)` removed.

server/webhook.go
```diff
@@ -184,9 +194,7 @@ func (p *Plugin) handleWebhook(w http.ResponseWriter, r *http.Request) {
 					continue
 				}
 				if info.Settings.Notifications {
-					if err := p.CreateBotDMPost(userTo, res.Message, "custom_git_review_request"); err != nil {
-						p.client.Log.Warn("can't send dm post", "err", err.Error())
-					}
+					p.sendDMNotification(userTo, res.Message)
```
nang2049, line 197: Worth confirming scope as this dedups the `res.ToUsers` DMs but the `res.ToChannels` loop still calls `p.client.Post.CreatePost` directly with no claim. The QA notes repro is two webhook entries pointing at the same endpoint so a subscribed channel will still get double posts after this change.

The ticket is DM scoped, so this may be deliberate and If it is meant to be covered here the same claim around the `ToChannels` post would do it.

nang2049 (commented):
Thanks @avasconcelos114

nang2049 (approved):

ogi-m (approved):
LGTM! 

Tested with bot DM notifications and subscriptions, events are posted only once

PR #703, opened 2026-08-19
Persist the GitLab URL from the setup flow to the plugin configuration

### LINKED ISSUE #37497: [Bug]: Gitlab custom URL it is not respected when using setup
### Before you file a bug report

- [x] I have checked the [issue tracker](https://github.com/mattermost/mattermost/issues) and have not found an issue that matches the one I'm filing.
- [x] This issue is not a troubleshooting question. Troubleshooting questions go here: https://forum.mattermost.com/c/trouble-shoot/16.
- [x] This issue is not a feature request. You can request features and make product suggestions here: https://mattermost.uservoice.com/forums/306457-general/.
- [x] This issue reproduces on one of the [currently supported server versions](https://docs.mattermost.com/about/mattermost-server-releases.html#latest-releases).
- [x] I have read the [contribution guidelines](https://github.com/mattermost/mattermost/blob/master/CONTRIBUTING.md) and the [Mattermost Handbook Contribution Guidelines](https://handbook.mattermost.com/contributors/contributors/guidelines/contribution-guidelines).

### Mattermost Server Version

Server Version: 11.8.3 Database Schema Version: 184 Build Number: 28791906131 Database: postgres

### Operating System

All

### Steps to reproduce

Set up GitLab Integration using GitLab /setup and select a custom GitLab URL. After completing the configuration, clicking Connect redirects to gitlab.com instead of the custom GitLab URL provided during setup.

### Expected behavior

The connection uses the custom URL

### Observed behavior

There was an error connecting to your GitLab: unable to connect user to GitLab: GET https://gitlab.com/api/v4/user: 401 {message: 401 Unauthorized} Please double check your configuration.

Instead of git.example.com

### Log Output

```shell

```

### Additional Information

I solve it doing the setup in System Console

ahmdyaasiin:
#### Summary

Running `/setup` and choosing a custom (self-hosted) GitLab URL stored the URL only on the in-memory `FlowManager` and in the KV instance configuration — it never reached the plugin settings. The GitLab API client is rebuilt from `config.GitlabURL` in `OnConfigurationChange`, so it stayed pointed at `https://gitlab.com`.

As a result, the OAuth authorize/token steps went to the self-hosted instance (through the KV instance credentials), but the follow-up `GET /api/v4/user` was sent to gitlab.com with the self-hosted token and failed:

> unable to connect user to GitLab: GET https://gitlab.com/api/v4/user: 401

This also matches the reporter's workaround: configuring the URL through the System Console (which does write the plugin setting) makes the connection work.

Changes:

- `submitGitlabURL` now persists the submitted URL through `setGitlabURL` instead of only assigning the in-memory field.
- `setGitlabURL` writes the URL into the saved plugin configuration — previously it saved the configuration map untouched (leftover TODO). The configuration is cloned first, following the concurrency strategy documented in `configuration.go`.
- Added a unit test asserting the URL reaches the saved plugin settings.

`make test` passes (382 tests).

#### Ticket Link

Fixes: https://github.com/mattermost/mattermost/issues/37497

## Change Impact: 🟡 Medium

**Reasoning:** The change updates plugin configuration persistence in the `/setup` flow. It affects a user-facing GitLab connection path but remains isolated to the flow manager and its tests.

**Regression Risk:** The risk is moderate because the change alters configuration saving behavior. Unit tests cover successful saves and save failures. Manual validation should confirm both hosted and self-hosted GitLab setup flows.

** QA Recommendation:** Perform focused manual QA for custom GitLab URLs and verify API requests use the saved URL. Skipping manual QA carries a moderate risk of missing integration-specific configuration issues.

*Generated by CodeRabbitAI*

### DIFF
--- a/server/flow.go
+++ b/server/flow.go
@@ -806,8 +806,6 @@ func (fm *FlowManager) submitChannelAnnouncement(f *flow.Flow, submitted map[str
 }
 
 func (fm *FlowManager) setGitlabURL(gitlabURL string) error {
-	fm.gitlabURL = gitlabURL
-
 	config := fm.getConfiguration().Clone()
 	config.GitlabURL = gitlabURL
 
@@ -820,6 +818,8 @@ func (fm *FlowManager) setGitlabURL(gitlabURL string) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to save plugin config")
 	}
+
+	fm.gitlabURL = gitlabURL
 
 	return nil
 }
--- a/server/flow.go
+++ b/server/flow.go
@@ -421,7 +421,9 @@ func (fm *FlowManager) submitGitlabURL(f *flow.Flow, submitted map[string]any) (
 		return "", nil, errorList, nil
 	}
 
-	fm.gitlabURL = gitlabURL
+	if err := fm.setGitlabURL(gitlabURL); err != nil {
+		return "", nil, nil, errors.Wrap(err, "failed to save GitLab URL")
+	}
 
 	return "", flow.State{
 		keyGitlabURL: gitlabURL,
@@ -806,8 +808,8 @@ func (fm *FlowManager) submitChannelAnnouncement(f *flow.Flow, submitted map[str
 func (fm *FlowManager) setGitlabURL(gitlabURL string) error {
 	fm.gitlabURL = gitlabURL
 
-	// will need to get gitlab url from plugin config
-	config := fm.getConfiguration()
+	config := fm.getConfiguration().Clone()
+	config.GitlabURL = gitlabURL
 
 	configMap, err := config.ToMap()
 	if err != nil {
--- a/server/flow_test.go
+++ b/server/flow_test.go
@@ -4,8 +4,10 @@
 package main
 
 import (
+	"net/http"
 	"testing"
 
+	"github.com/mattermost/mattermost/server/public/model"
 	"github.com/mattermost/mattermost/server/public/plugin/plugintest"
 	"github.com/mattermost/mattermost/server/public/pluginapi"
 	"github.com/stretchr/testify/assert"
@@ -38,6 +40,27 @@ func TestSetGitlabURL(t *testing.T) {
 	assert.Equal(t, "https://git.example.com", savedConfig["gitlaburl"])
 
 	assert.Equal(t, "https://gitlab.com", config.GitlabURL)
+
+	api.AssertExpectations(t)
+}
+
+func TestSetGitlabURLSaveFailure(t *testing.T) {
+	config := &configuration{
+		GitlabURL: "https://gitlab.com",
+	}
+
+	api := &plugintest.API{}
+	api.On("SavePluginConfig", mock.Anything).Return(model.NewAppError("SavePluginConfig", "app.plugin.config.app_error", nil, "", http.StatusInternalServerError)).Once()
+
+	fm := &FlowManager{
+		client:           pluginapi.NewClient(api, nil),
+		getConfiguration: func() *configuration { return config },
+	}
+
+	err := fm.setGitlabURL("https://git.example.com")
+	require.Error(t, err)
+
+	assert.Empty(t, fm.gitlabURL)
 
 	api.AssertExpectations(t)
 }
--- a/server/flow_test.go
+++ b/server/flow_test.go
@@ -1,0 +1,43 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package main
+
+import (
+	"testing"
+
+	"github.com/mattermost/mattermost/server/public/plugin/plugintest"
+	"github.com/mattermost/mattermost/server/public/pluginapi"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSetGitlabURL(t *testing.T) {
+	config := &configuration{
+		GitlabURL: "https://gitlab.com",
+	}
+
+	var savedConfig map[string]any
+	api := &plugintest.API{}
+	api.On("SavePluginConfig", mock.Anything).Run(func(args mock.Arguments) {
+		savedConfig = args.Get(0).(map[string]any)
+	}).Return(nil).Once()
+
+	fm := &FlowManager{
+		client:           pluginapi.NewClient(api, nil),
+		getConfiguration: func() *configuration { return config },
+	}
+
+	err := fm.setGitlabURL("https://git.example.com")
+	require.NoError(t, err)
+
+	assert.Equal(t, "https://git.example.com", fm.gitlabURL)
+
+	require.NotNil(t, savedConfig)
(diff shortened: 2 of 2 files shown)


server/flow.go
```diff
@@ -421,7 +421,9 @@ func (fm *FlowManager) submitGitlabURL(f *flow.Flow, submitted map[string]any) (
 		return "", nil, errorList, nil
 	}
 
-	fm.gitlabURL = gitlabURL
+	if err := fm.setGitlabURL(gitlabURL); err != nil {
```
nang2049: This is now a privilege escalation path, I think.

`submitDelegateSelection` (line 333) doesn't check that the delegate is a system admin, and the comment above it acknowledges non-admin delegates are expected. So a delegated non-admin walks the full setup flow, including `stepInstanceURL`, and reaches this submit handler.

Before this change that was harmless: `submitGitlabURL` only set the in-memory `fm.gitlabURL` and the KV instance config. Now it persists to the plugin configuration via `SavePluginConfig`, which triggers `OnConfigurationChange` and rebuilds `p.GitlabClient` for the whole plugin. A non-admin delegate can therefore point every subsequent GitLab API call at a host they control.

`stepSetDefaultInstance` already guards against exactly this at line 834:

```go
isAdmin, err := fm.isAuthorizedSysAdmin(f.UserID)
if err != nil {
	return "", nil, fmt.Errorf("failed to verify admin privileges: %w", err)
}
if !isAdmin {
	return "", nil, fmt.Errorf("only system administrators can set the default instance")
}
```

Same check here would do it, or verify the delegate is an admin in `submitDelegateSelection` so the whole flow is admin-only. Worth noting the `Yes` button at line 373 calls `setGitlabURL(gitlab.Gitlabdotcom)` and needs the same treatment.

server/flow.go
```diff
@@ -819,6 +819,8 @@ func (fm *FlowManager) setGitlabURL(gitlabURL string) error {
 		return errors.Wrap(err, "failed to save plugin config")
 	}
 
+	fm.gitlabURL = gitlabURL
```
nang2049: `fm.gitlabURL` keeps the raw submitted value while the persisted config gets normalized: `sanitize()` strips trailing slashes on the next `OnConfigurationChange`, which this save triggers.

So for `https://git.example.com/` the plugin config ends up as `https://git.example.com` but `fm.gitlabURL` keeps the slash, and it's used verbatim for `InstanceConfiguration.GitlabURL` at line 542. The flow state at line 429 has the same issue, and it renders into the OAuth instructions as `https://git.example.com//-/profile/applications`.

Trimming before assigning, or assigning `fm.getConfiguration().GitlabURL` after the save, keeps the two in sync.

server/flow_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	fm := &FlowManager{
+		client:           pluginapi.NewClient(api, nil),
+		getConfiguration: func() *configuration { return config },
+	}
+
+	err := fm.setGitlabURL("https://git.example.com")
+	require.Error(t, err)
+
+	assert.Empty(t, fm.gitlabURL)
```
nang2049: This passes trivially. `fm.gitlabURL` is never set in the fixture, so it's empty either way and the assertion holds even if the assignment were still above the save.

Seeding it first is what actually pins the behavior:

```go
fm := &FlowManager{
	client:           pluginapi.NewClient(api, nil),
	getConfiguration: func() *configuration { return config },
	gitlabURL:        "https://gitlab.com",
}

// ...
assert.Equal(t, "https://gitlab.com", fm.gitlabURL)
```

server/flow_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/mattermost/mattermost/server/public/model"
+	"github.com/mattermost/mattermost/server/public/plugin/plugintest"
+	"github.com/mattermost/mattermost/server/public/pluginapi"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSetGitlabURL(t *testing.T) {
```
nang2049: Both tests cover `setGitlabURL`, which already existed. Nothing covers `submitGitlabURL`, which is the function the bug was actually in and the one this PR changed. A case asserting it persists the URL, and that a `SavePluginConfig` failure surfaces as an error rather than a silent success, would keep the regression from coming back.

server/flow.go
nang2049: I think this opens a privilege escalation path.

Neither `submitDelegateSelection` (line 333) nor `StartSetupWizard` (line 239) checks that the delegate is a system admin, and the comment above `submitDelegateSelection` acknowledges non-admin delegates are expected. The delegate is started on the full setup flow from the first step, so they reach `stepInstanceURL` and this handler.

What changes here is the blast radius. Before, a non-admin delegate could write an arbitrary GitLab URL into the KV instance config via `installInstance`, but `p.GitlabClient` is built from the *plugin* config in `OnConfigurationChange` (`configuration.go:243`), so the plugin's own API client was unaffected. Now `submitGitlabURL` persists through `SavePluginConfig`, which fires `OnConfigurationChange` and rebuilds that client — so a non-admin delegate can repoint every subsequent GitLab API call at a host they control.

`stepSetDefaultInstance` already guards this exact case at line 834:

```go
isAdmin, err := fm.isAuthorizedSysAdmin(f.UserID)
if err != nil {
	return "", nil, fmt.Errorf("failed to verify admin privileges: %w", err)
}
if !isAdmin {
	return "", nil, fmt.Errorf("only system administrators can set the default instance")
}
```

Same check here would cover it, or verify the delegate is an admin in `submitDelegateSelection` so the whole flow is admin-only. The `Yes` button at line 373 calls `setGitlabURL(gitlab.Gitlabdotcom)` and needs the same treatment.

server/flow_test.go
nang2049: Both tests target `setGitlabURL`, which already existed and already saved the config. The behavior this PR actually changes is in `submitGitlabURL`, and nothing covers it.

It's cheap to add — `f` is unused in the body, so `nil` works:

```go
nextStep, state, errorList, err := fm.submitGitlabURL(nil, map[string]any{"gitlab_url": "https://git.example.com"})
```

Then assert the URL reached `savedConfig["gitlaburl"]`, and that a `SavePluginConfig` failure comes back as an error rather than a silent success. That's the regression worth locking in, since the old code returned successfully while never persisting anything.

server/flow.go
nang2049, line 424: `submitDelegateSelection` doesn't check the delegate is an admin so a non admin can reach this step. That was fine when it only set the in memory URL but now it writes the plugin setting and rebuilds `p.GitlabClient` so they could point the whole integration at their own host.

`stepSetDefaultInstance` does an `isAuthorizedSysAdmin(f.UserID)` check for the same reason. Can we add that here and on the gitlab.com button above?

server/flow_test.go
nang2049, line 18: Both tests cover `setGitlabURL`, which already saved the config. Nothing covers `submitGitlabURL`, which is what actually changed. `f` is unused in the body so you can pass nil.

nang2049 (changes requested):
Thanks @ahmdyaasiin just a couple of comments :)

PR #714, opened 2026-09-21
MM-70733 Ensuring configuration detection includes installed instances

avasconcelos114:
#### Summary

In the current version of the plugin, if the client id and secret in the global configuration are missing, the plugin assumed the setup isn't completed and will not process the instance-specific commands. This PR addresses that by adding a layer of config and instance detection so that only real cases of missing configuration are treated as such, and users can rely exclusively on the instance-based setups in order to use the plugin normally

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-70733

## Change Impact: 🔴 High

**Reasoning:** Changes affect GitLab authentication, instance configuration, KV-backed persistence, OAuth flows, and command execution across multiple modules. These changes affect critical user-facing paths.

**Regression Risk:** Medium to high. Configuration resolution and OAuth behavior changed, although regression tests cover several configured and unconfigured cases.

**QA Recommendation:** Perform focused manual QA for installation, instance setup and removal, OAuth connection, default-instance changes, command execution, autocomplete, and legacy configuration. Skipping manual QA carries a high risk of missing configuration or authentication regressions.

*Generated by CodeRabbitAI*


server/api.go
```diff
@@ -168,9 +168,7 @@ func (p *Plugin) withRecovery(next http.Handler) http.Handler {
 
 func (p *Plugin) checkConfigured(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		config := p.getConfiguration()
-
-		if err := config.IsValid(); err != nil {
+		if err := p.isConfigured(); err != nil {
```
nang2049, line 171: `checkConfigured` wraps every `/api/v1` route so each call to `/connected`  and `/lhs-data` now does two KV reads: the name list and the config map. `getConnected` resolves again afterwards and`/connected can do four KV reads.

Suggest to cache the resolved `effectiveConfig` in memory and refresh it when you call `refreshGitlabClient` (`OnConfigurationChange`, install and uninstall). Hot paths then read the cache instead of KV.

server/plugin.go
```diff
@@ -285,53 +291,89 @@ func (p *Plugin) getOAuthConfig() (*oauth2.Config, error) {
...
-		clientID = instanceConfig.GitlabOAuthClientID
-		clientSecret = instanceConfig.GitlabOAuthClientSecret
-		rawURL = instanceConfig.GitlabURL
+		effective = effectiveConfig{
+			InstanceName: config.DefaultInstanceName,
+			GitlabURL:    instanceConfig.GitlabURL,
+			ClientID:     instanceConfig.GitlabOAuthClientID,
+			ClientSecret: instanceConfig.GitlabOAuthClientSecret,
+		}
 	case config.GitlabOAuthClientID != "" && config.GitlabOAuthClientSecret != "" && config.GitlabURL != "":
```
nang2049, line 340: Could `getInstance` return a sentinel error for the "does not exist" and "not found" cases? Then fall back to legacy only on that error and propagate everything else.

server/plugin.go
```diff
@@ -285,53 +291,89 @@ func (p *Plugin) getOAuthConfig() (*oauth2.Config, error) {
...
+	if config.UsePreregisteredApplication {
+		return &effectiveConfig{GitlabURL: config.GitlabURL}
 	}
 
-	return clientID, clientSecret, gitlabURL, nil
+	if effective, err := p.resolveEffectiveConfig(config); err == nil {
+		return effective
+	}
+
+	return &effectiveConfig{
```
nang2049, line 373: This fallback returns `config.GitlabURL`, which defaults to `https://gitlab.com`, not empty. So when `refreshGitlabClient` runs during a failed or transient resolution, a working self-hosted client gets replaced by one pointing at gitlab.com (same finding as CodeRabbit)

server/configuration.go
```diff
@@ -240,11 +230,21 @@ func (p *Plugin) OnConfigurationChange() error {
...
 }
 
+// refreshGitlabClient rebuilds the GitLab API client using the currently effective GitLab URL,
+// which may come from the KV-backed default instance rather than legacy plugin settings. Call
+// this after any change that could affect the effective instance (plugin configuration changes,
+// or direct KV mutations such as installing/uninstalling an instance).
+func (p *Plugin) refreshGitlabClient() {
+	config := p.getConfiguration()
+	effective := p.resolveEffectiveConfigOrDefault(config)
+	p.GitlabClient = gitlab.New(effective.GitlabURL, config.GitlabGroup, p.isNamespaceAllowed)
```
nang2049, line 245: `p.GitlabClient` is assigned without any synchronization. Before this PR only `OnConfigurationChange` wrote it. Now `installInstance` and `uninstallInstance` write it too, and those run on slash-command and flow request goroutines while HTTP handlers and webhooks read `p.GitlabClient` at the same time. That's a data race `-race` would catch under a real workload.

Could we put it behind a getter/setter using a mutex or `atomic.Pointer`?

server/flow.go
```diff
@@ -837,15 +815,18 @@ func (fm *FlowManager) stepSetDefaultInstance() flow.Step {
...
 		}).
 		WithButton(flow.Button{
-			Name:    "No",
-			Color:   flow.ColorDefault,
-			OnClick: flow.Goto(stepDone),
+			Name:  "No",
+			Color: flow.ColorDefault,
+			// Declining to set the default instance still leaves it configured; continue the
+			// wizard so the admin can connect their account and set up the webhook.
+			OnClick: flow.Goto(stepOAuthConnect),
```
nang2049, line 830: +1 to CodeRabbit. If another instance is already the default and the admin picks "No", `/oauth/connect` resolves `DefaultInstanceName`.

One more thing to keep in mind for the fix: before this change "No" went to `stepDone`, which isn't registered in `oauthFlow`. So `/gitlab setup oauth` followed by "No" was already broken, and whatever we do here has to work in both `setupFlow` and `oauthFlow`.

server/flow.go
```diff
@@ -228,11 +229,12 @@ func continueButton(next flow.Name) flow.Button {
...
 func (fm *FlowManager) getBaseState() flow.State {
 	config := fm.getConfiguration()
-	isOAuthConfigured := config.GitlabOAuthClientID != "" || config.GitlabOAuthClientSecret != ""
 	return flow.State{
 		keyGitlabURL:                   config.GitlabURL,
 		keyUsePreregisteredApplication: config.UsePreregisteredApplication,
-		keyIsOAuthConfigured:           isOAuthConfigured,
+		// Reflects whether a GitLab instance (KV-backed or legacy plugin settings) is already
+		// configured, so the wizard can warn that it will be overwritten.
+		keyIsOAuthConfigured: fm.canConnect(),
```
nang2049, line 237: This drives the "**Any existing OAuth configuration will be overwritten.**" text in the OAuth dialog. With KV instances, submitting that dialog creates a new instance, and `installInstance` rejects duplicate names so nothing actually gets overwritten. Switching to `canConnect()` also shows the warning in more cases. Could we reword it (e.g. "This will add a new instance")?

server/command.go
```diff
@@ -1095,7 +1094,7 @@ func (p *Plugin) isAuthorizedSysAdmin(userID string) (bool, error) {
 }
 
 func (p *Plugin) getAutocompleteData(config *configuration) *model.AutocompleteData {
-	if !config.IsOAuthConfigured() {
+	if p.isConfigured() != nil {
```
nang2049, line 1097: Autocomplete is only rebuilt in `OnConfigurationChange`. `uninstallInstance` only changes KV, so after removing the default or last instance the full command list stays registered. Consider re-registering the slash command in the same place you refresh the client.

server/command_test.go
```diff
@@ -896,3 +897,53 @@ func TestInstanceCommands(t *testing.T) {
 		})
 	})
 }
+
+// TestExecuteCommandRecognizesKVOnlyInstance guards against the regression fixed by this
```
nang2049: Nit: "guards against the regression fixed by this change" and "see #595" describe PR history and will read oddly once this is merged. The test name already says what it checks, so I'd drop these doc comments (same for the one at L926).

nang2049 (commented):
Thanks @avasconcelos114

